Issue #3, opened 2016-10-02
Add a test suit


khaledhosny:
We need a test suit of some kind, otherwise one is stabbing in the dark with such an ancient and widely used code base.

khaledhosny:
@adrientetar suggested I use the open source Adobe fonts, and I think that is the simplest way. I already used Source Serif to test the 64-bit changes, and I now need to figure out how to make and automated test suite out of it.

khaledhosny:
We have a test suite of sorts now.

miguelsousa:
I've discussed some ideas with @readroberts and we agreed that it'd be better to have some bite-size tests. So I'm going to look into building up such battery of tests.

khaledhosny:
Sounds great, thanks Miguel.

PR #11, opened 2017-06-06
make package installable and distributable as wheel

anthrotype:
PTAL @khaledhosny


khaledhosny:
Looks great, thanks!

Few general comments:
* I’d simply rename `python/` to `psautohint/`, since there is nothing else in `python/`. (we can later rename `source/` to `libpsautohint/` or so).
* Now I need to do `make PIP_OPTIONS=--user check` locally instead of just `make check` I know I should use virtualenv, but I find using them to be clumsy), what about swapping the default?
* A nitpick: I don’t like too long lines in commit messages, especially the first line. I think the most common format is using a 72-character column (I actually just use whatever vim’s default is 😄). So, for example, instead of:
  ```
  add setup.cfg for specifying universal wheel, the path to the LICENSE file, and zip format for sdist
  ```
  I’d use:
  ```
  Add setup.cfg

  For specifying universal wheel, the path to the LICENSE file, and zip
  format for sdist.
  ```

anthrotype:
it's still good to keep the package source in a different sub-directory (e.g. src or Lib or whatever), otherwise when you import `psautohint`, since the current directory is always in the PYTHONPATH, you risk inadvertently importing the source directory rather than the installed distribution. It's better to test the installed package than the one in the source tree, to prevent issues with packaging.

OK, I'll swap the default for PIP_OPTIONS and make the lines in my commit message shorter.

khaledhosny:
> it's still good to keep the package source in a different sub-directory (e.g. src or Lib or whatever), otherwise when you import psautohint, since the current directory is always in the PYTHONPATH, you risk inadvertently importing the source directory rather than the installed distribution. It's better to test the installed package than the one in the source tree, to prevent issues with packaging.

I see, makes sense.

> OK, I'll swap the default for PIP_OPTIONS and make the lines in my commit message shorter.

Scratch that. I found https://gist.github.com/codysoyland/2198913 and it seems to be a good enough way to manage virtual env, so keeping things as they are now should be fine.

khaledhosny:
Another issue, it seems `make check` will continue using the last version of the module unless the version number is increased, since `python -m pip install -r requirements.txt -v  .` does not force updating the module. This is not an issue with CI, but locally I’ll be testing older version and entirely miss any issues.

anthrotype:
maybe it would be better to use `-e .` as in editable?

.travis.yml
```diff
@@ -1,11 +1,16 @@
...
 
-os:
-  - linux
-  - osx
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.6
+    # The 'osx' environment uses the python2.7 by default.
+    # TODO: maybe also tests python3 on macOS
```
khaledhosny: Testing the default Python on macOS is good enough. We test on macOS just to make sure the C code builds there.

tests/Makefile
```diff
@@ -23,7 +23,7 @@ all: $(TESTS)
 %.tst: %.ufo FORCE
 	@echo "	Testing $<"
 	@rm -rf $@
-	@$(PYTHON) $(PYT_DIR)/autohint.py -all -qq -o $@ $<
+	@$(PYTHON) -m psautohint.autohint -all -qq -o $@ $<
```
khaledhosny: Since `PYT_DIR` is unused now, you should remove its declaration above.

setup.py
```diff
@@ -35,8 +36,36 @@
...
         )
 
-setup (name = "PsAutoHint",
-       version = "1.0",
-       description = "Python wrapper for Adobe's PostScrupt autohinter",
-       ext_modules = [module1])
-
+setup(name="psautohint",
+      version="1.1.0.dev0",
+      description="Python wrapper for Adobe's PostScrupt autohinter",
```
khaledhosny: Typo “PostScrupt” (OK it is *my* typo, but we might as well fix it now).

khaledhosny:
I think editable should be good, or use `--upgrade` and install the requirements in a separate line without `--upgrade`.

anthrotype:
I made the commit message shorter and removed the unused makefile vars.
I haven't added -e or --upgrade though.. What about for now you override those using the PIP_OPTIONS env var?
Going forward, I am actually thinking of trying to replace GNU make with a python-based build/test setup, so that will work on all the platforms, Windows included.

anthrotype:
Another alternative to requiring first to install before running the tests would be to have the `check` target depend on the `build` one, and then export the `PYTHONPATH` environment variable pointing to the built package inside `build/lib.*` directory just before running the tests.

This way one can do `make check` without worrying about sudo or virtual environment, because `check` only depends on `build`, not on `install`.

We run a python snippet that queries the sys module for the path to the platform-specific build directory (in the same way distutils/setuptools do):

```diff
diff --git a/Makefile b/Makefile
index da779faf..e30ad019 100644
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,21 @@ PYTHON ?= python
 # PIP_OPTIONS ?= --user
 PIP_OPTIONS ?=
 
+# Get the platform/version-specific build/lib.* folder
+define GET_BUILD_DIR
+import os
+import sys
+import sysconfig
+root = sys.argv[1]
+plat_lib_name = "lib.{platform}-{version[0]}.{version[1]}".format(
+    platform=sysconfig.get_platform(), version=sys.version_info)
+build_dir = os.path.abspath(os.path.join(root, "build", plat_lib_name))
+print(build_dir)
+endef
+
+# Extend sys.path with ./build/lib.* directory
+PYTHONPATH ?= $(shell $(PYTHON) -c '$(GET_BUILD_DIR)' $(ROOT_DIR))
+
 build:
        $(PYTHON) setup.py build
 
@@ -22,8 +37,8 @@ clean:
        $(PYTHON) setup.py clean --all
        make -C $(SRC_DIR) clean
 
-check: install
-       make -C $(TST_DIR)
+check: build
+       make -C $(TST_DIR) PYTHONPATH="$(PYTHONPATH)"
 
 format:
        clang-format -i `find $(SRC_DIR) -name '*.c'`
```

khaledhosny:
> Another alternative to requiring first to install before running the tests would be to have the check target depend on the build one, and then export the PYTHONPATH environment variable pointing to the built package inside build/lib.* directory just before running the tests.

Even better!

khaledhosny:
We do need a real test harness of course, see #3 and #6.

khaledhosny:
All good now, please let me know when you are done with this.

khaledhosny:
Can we move `source/_psautohint.c` to `python/psautohint/_psautohint.c` as well?

anthrotype:
ok i'm done for this round, feel free to merge, thanks.

khaledhosny:
Thanks Cosimo!

PR #17, opened 2017-11-27
disabling gcc's tree vectorize fixes segfault

anthrotype:
Fixes https://github.com/khaledhosny/psautohint/issues/16

### DIFF
--- a/libpsautohint/Makefile
+++ b/libpsautohint/Makefile
@@ -23,6 +23,7 @@ CFLAGS = \
 	-Wunused-macros \
 	-Wmissing-prototypes \
 	-O3 \
+	-fno-tree-vectorize \
 	$(NULL)
 
 ifeq ($(firstword $(shell $(CC) --version)),clang)
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,22 @@
 import platform
 from setuptools import setup, Extension
+from setuptools.command.build_ext import build_ext
+
+
+class CustomBuildExt(build_ext):
+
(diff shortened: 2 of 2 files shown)


anthrotype:
should we also add the `-fno-tree-vectorize` flag to CFLAGS in the libpsautohint/Makefile?

khaledhosny:
Thanks!

anthrotype:
I wonder if compiling with clang instead of gcc causes the same issue.

anthrotype:
yes, I confirm clang doesn't segfault with -O3 (I tried both clang-3.8 and clang-3.9).
I run `CC=clang-3.9 python setup.py build` at the commit immediately before this PR.

the problem is distutils can only check if compiler is "unix", can't distinguish whether it's gcc or clang.
Anyway, I don't think it's a problem if tree-vectorize optimization is disabled.

behdad:
It's either a gcc bug, or undefined-behavior in our code.  This is just a workaround.

khaledhosny:
My thoughts exactly. I was planning to see if I can build with ubsan and see if it reports anything.

anthrotype:
I recompiled with CFLAGS="-fsanitize=undefined" LDFLAGS="-fsanitize=undefined" and when I run make check I get hundreds of runtime errors like this

libpsautohint/src/auto.c:66:14: runtime error: member access within misaligned address 0x7fda19c9c974 for type 'struct _clrval', which requires 8 byte alignment

PR #20, opened 2018-02-12
Python files cleanup

miguelsousa:


python/psautohint/fdTools.py
```diff
@@ -151,7 +151,7 @@ def buildBlueLists(self):
             for key in keyList:
                 if key.endswith("Overshoot"):
                     width = eval("self.%s" % (key))
-                    if width is None:
+                    if width is not None:
```
khaledhosny: I’d rather prefer if this commit was squashed into 12629a6, as I prefer keeping things `git bisect`-able as much as possible

PR #21, opened 2018-02-12
fix Python shebang

jbicha:
The standard shebang for Python if you want to allow users to use whatever Python is in their path is
`#!/usr/bin/env python'

I noticed this issue while packaging psautohint for [Debian](https://lintian.debian.org/tags/script-uses-bin-env.html) .


khaledhosny:
We can as well just drop the shebang here, these scripts are not called directly.

jbicha:
That's fine with me. Do you want to do that and close this pull request?

khaledhosny:
I just did.

PR #26, opened 2018-02-14
More fixes to python files

miguelsousa:


python/psautohint/otfFont.py
```diff
@@ -820,7 +821,7 @@ def convertBezToT2(bezString):
                 if int(val1) == val2:
                     argList.append(val2)
                 else:
-                    argList.append("% 100 div" % (str(int(val1 * 100))))
+                    argList.append("%s 100 div" % str(int(val1 * 100)))
```
cjchapman: Aren't we standardizing on `x.format(y)` instead of ` x % y`?

miguelsousa: Preferably yes. This time I just fixed what was broken. I can take a look at all the lines that use string replacement in a later round.

cjchapman: OK.

PR #30, opened 2018-02-21
Replace evals

miguelsousa:


python/psautohint/ufoFont.py
```diff
@@ -512,7 +513,7 @@ def readHashMap(self):
         if os.path.exists(hashPath):
             with open(hashPath, "rt") as fp:
                 data = fp.read()
-            newMap = eval(data)
+            newMap = ast.literal_eval(data)
```
cjchapman: What is `data` and what is `newMap`?
(I ask because I'm wondering if it is really necessary to use even a `literal_eval` here.)

python/psautohint/ufoFont.py
```diff
@@ -1103,9 +1104,8 @@ def buildGlyphHashValue(self, width, outlineXML, glyphName,
                                          "yScale", "xOffset", "yOffset"]:
                         try:
                             value = childContour.attrib[transformTag]
-                            rval = eval(value)
-                            if int(rval) == rval:
-                                value = str(int(rval))
+                            if self._represents_integer(value):
```
cjchapman: The `_represents_integer()` method seems like overkill. I think you can just do `value.isdigit()`.

I'm confused by the next line. If you know it's a string that represents an integer, why are you converting it to an integer and then back into a string?

python/psautohint/ufoFont.py
```diff
@@ -1122,6 +1122,14 @@ def buildGlyphHashValue(self, width, outlineXML, glyphName,
             glyph_hash = hashlib.sha512(data.encode("utf-8")).hexdigest()
         return glyph_hash, dataList
 
+    @staticmethod
+    def _represents_integer(string):
```
cjchapman: I don't think this is needed. (See comment above.)

python/psautohint/ufoFont.py
miguelsousa: I considered `isdigit()`,  but it can't be used for validating an integer because it returns `False` for `+1`, `1.0`, `-1` (and perhaps others I haven't even thought of).

python/psautohint/ufoFont.py
miguelsousa: I agree it seems overkill, but `isdigit()` is not a viable option.

python/psautohint/ufoFont.py
miguelsousa: `data` is a string and `newData` is a dictionary. (I alluded to it in the commit notes)

python/psautohint/ufoFont.py
cjchapman: OK... so do you still need the next line that converts a string to an integer back to a string?

miguelsousa: This is what I wrote in the commit notes,
> Not sure why integer-like string values need special handling here, but let's leave that for later.

PR #32, opened 2018-03-02
Multiple masters support

khaledhosny:
I think this is mergable now, any remaining issues can be fixed in master.


behdad:
Does this make CFF2 varfont hinting?

readroberts:
Not yet. That is the goal, and it is coming soon. However, the wip/mm branch was only gotten to the point where it runs and looks like it does CFF2 var font hinting. It has not yet been tested at all - no idea yet if hint substitution or stem3 hints are working.

behdad:
What was stem3 again? I don't remember that.

readroberts:
'stem3' was my attempt at short hand for hstem3 and vstem3, the counter control hints, used for things like m and 'w'.

behdad:
> 'stem3' was my attempt at short hand for hstem3 and vstem3, the counter control hints, used for things like m and 'w'.

Does this module still hint Type1?

miguelsousa:
> Does this module still hint Type1?

It should, but it's giving an error at the moment.
```
Hinting font font.pfa. Start time: Thu Mar  8 18:00:23 2018.
Font file must be a CFF or OTF fontfile: font.pfa.
ACFontError: Font file must be CFF or OTF file: font.pfa.
Error parsing font file <font.pfa>.
```

readroberts:
That's because Khaled took removed the Python code that converts Type1 fonts when he adapted psautohint.py from autohint.py. It would be trivial to restore this. To answer Behdad's question more fully: the libpsautohint.so module works with bez file data, an early simplified version text version of  Type 1, one bez file data string per glyph. The Python wrapper uses fonttools to open an OpenType/CFF file, and convert each CFF T2 charstring in turn to to bez data, calls the libpsautohint module, gets the updated bez file data for the glyph, and writes that back to the OpenType file.   If the source font is 'CFF', the Python wrapper creates an minimal TTFont, adds the CFF file as a CFF table, then proceeds with case 1. If the source font is a Type 1 font, then the Python wrapper calls the  AFDKO tool 'tx' to convert it to CFF, then goes to case 2.

miguelsousa:
@readroberts got it. I think it will be valuable to support Type 1 again, but I don't like the dependence on *tx*. We need to find a different method.

khaledhosny:
Yes, Type 1 support was removed to drop dependency on TX. I think it should
be possible to support Type 1 directly using FontTools similar to how CFF
is currently handled.

readroberts:
Yes, I agree. When autohint.py was written, fonttools didn't offer easy access to T1 fonts, but now it does.

miguelsousa:
[fonttools can open PFAs](https://github.com/fonttools/fonttools/blob/master/NEWS.rst#22-released-2008-05-18), and judging from the [t1Lib tests](https://github.com/fonttools/fonttools/blob/master/Tests/t1Lib/t1Lib_test.py) it can open PFBs as well,
so I'm guessing it should be possible to support it again via fontTools.

PR #37, opened 2018-04-04
Simplify memory handling

khaledhosny:


libpsautohint/autohintexe.c
```diff
@@ -401,25 +401,8 @@ main(int argc, char* argv[])
                 openReportFile(bezName, fileSuffix);
             }
 
-            result = AutoColorString(bezdata, fontinfo, output, &outputsize,
+            result = AutoColorString(bezdata, fontinfo, &output, &outputsize,
                                      allowEdit, allowHintSub, roundCoords, debug);
-            if (result == AC_DestBuffOfloError) {
```
cjchapman: I'm not sure if it's OK for this to fall through to the return or not or if you should still check result at this point.

libpsautohint/autohintexe.c
```diff
@@ -464,24 +447,8 @@ main(int argc, char* argv[])
             outGlyphs[i] = malloc(outputSizes[i]);
         }
 
-        result = AutoColorString(inGlyphs[0], fontinfo, outGlyphs[0], &outputSizes[0],
+        result = AutoColorString(inGlyphs[0], fontinfo, &outGlyphs[0], &outputSizes[0],
                                  allowEdit, allowHintSub, roundCoords, debug);
-        if (result == AC_DestBuffOfloError) {
```
cjchapman: I do think you should check result here before you assign another value to result just below.

libpsautohint/include/psautohint.h
```diff
@@ -134,13 +133,12 @@ ACLIB_API void AC_SetReportRetryCB(AC_RETRYPTR retryCB);
  * pointer to null terminated C string containing fontinfo for the bez glyph.
  *
  * Hint information is added to the bez data and returned to the caller through
- * the buffer dstbezdata. dstbezdata must be allocated before the call and a
+ * the buffer dstbezdata. dstbezdata should be allocated before the call and a
```
cjchapman: Why did you change the wording from `must` to `should` here? Is there a scenario in which it's OK to not allocate it before the call? If so, it might be helpful to describe that scenario.

khaledhosny: Since we now re-allocate the output buffer if it is too small, not allocating it and passing 0 for the length should work. But actually, after checking that, it doesn’t, I’ll revert the wording here.

PR #41, opened 2018-04-20
Use markdown for 'long_description'

anthrotype:
The new PyPI finally added support for Github Flavored Markdown as project's `long_description` format, in addition to reStructuredText.

The nice thing about it is that both Github and PyPI use the same library (`cmark`) to render the markdown documents to html, so if it renders on Github, you can be certain that it will also render on PyPI -- hence, no need to the extra checks.


anthrotype:
I converted it automatically with pandoc.
Here is how it renders on my fork:
https://github.com/anthrotype/psautohint/blob/markdown-description/README.md

PR #42, opened 2018-05-26
Use cibuildwheel to produce wheels

madig:
Hi there!
i'd like to have wheels of psautohint available so I don't have to install a compiler elsewhere.

For this PR, I use https://github.com/joerick/cibuildwheel to build wheels for 2.7 and 3.6 for Linux, macOS and Windows 32 and 64 bits. The wheels build, but there is no test and deploy step still.


anthrotype:
Nice! You should be able to run the tests with CIBW_TEST_COMMAND and CIBW_TEST_REQUIRES

anthrotype:
I also used cibuildwheel for uharfbuzz, maybe you can find that useful https://github.com/trufont/uharfbuzz/blob/master/.travis.yml

madig:
Ugh. The test setup is a bit... special.

madig:
@khaledhosny @miguelsousa if you're working on this right now, maybe have a look at cibuildwheel :)

madig:
Oh boy, you might want to trim your test regime. 30+ Minutes for a full build is hard.

anthrotype:
why it takes so long?

.appveyor.yml
```diff
@@ -1,21 +1,18 @@
...
-install:
-  - git submodule update --init
-  - python -m pip install --upgrade pip
-  - pip install -r requirements.txt
-  - pip install -r dev-requirements.txt
+build_script:
+  - pip install cibuildwheel==0.9.2 coverage codecov
+  - cibuildwheel --output-dir wheelhouse
+  - coverage combine tests
+  - codecov
```
anthrotype: @madig I think you can keep calling codecov via tox, as it was before `tox -e codecov`

.appveyor.yml
```diff
@@ -1,21 +1,18 @@
 environment:
-  matrix:
-    - TOXENV: py27-cov
-    - TOXENV: py36-cov
+  global:
+    CIBW_SKIP: cp33-* cp34-* cp35-*
+    CIBW_TEST_REQUIRES: pytest pytest-cov
+    CIBW_TEST_COMMAND: "cd {project}\\tests && pytest --cov"
```
anthrotype: maybe you could have `tox` as the only `CIBW_TEST_REQUIRES`, and again `tox` as the only `CIBW_TEST_COMMAND`. This way the CI and the local developer's test workflow is the same and we avoid duplication.

.coverage
```diff
@@ -0,0 +1 @@
...360],[360,361],[361,362],[362,363],[363,364],[364,365],[365,366],[366,367],[367,368],[368,370],[370,371],[371,373],[373,374],[374,377],[-377,377],[377,378],[378,-377],[377,381],[-381,381],[381,382],[382,-381],[381,385],[-385,385],[385,386],[386,424],[424,433],[433,441],[441,445],[445,455],[455,462],[462,468],[468,491],[491,506],[506,511],[511,534],[534,556],[556,567],[567,588],[588,595],[595,608],[608,633],[633,701],[701,719],[719,756],[756,761],[761,807],[807,813],[813,843],[843,861],[861,980],[980,985],[985,1019],[1019,1029],[1029,1125],[1125,1133],[1133,1175],[1175,1180],[1180,1186],[1186,1196],[1196,-385],[385,1202],[1202,1215],[1215,1297],[-1297,1297],[1297,1298],[1298,1299],[1299,1301],[1301,1338],[1338,1361],[1361,-1297],[1297,1373],[1373,1593],[1593,1606],[-1606,1606],[1606,1609],[1609,1620],[1620,-1606],[1606,1649],[1649,1670],[1670,1692],[1692,1699],[1699,1722],[1722,1723],[1723,1724],[1724,1725],[1725,1726],[1726,1727],[1727,1728],[1728,1729],[1729,1730],[1730,1731],[1731,1732],[1732,1733],[1733,1734],[1734,1738],[1738,1748],[1748,2138],[2138,2160],[2160,2176],[2176,-113]],"C:\\UsersLocal\\nikolaus.waxweiler\\Development\\psautohint\\python\\psautohint\\__main__.py":[]}}
```
anthrotype: i think you didn't want to commit this `.coverage` file. Should be .gitignored

.appveyor.yml
madig: Hm... will it pick up the built package?

miguelsousa:
the tests aren't fast, but it seems that cibuildwheel is building too many wheels. Here's the list I collected from the log,
```
New WHEEL info tags: cp27-cp27m-manylinux1_x86_64
New WHEEL info tags: cp27-cp27mu-manylinux1_x86_64
New WHEEL info tags: cp36-cp36m-manylinux1_x86_64
New WHEEL info tags: cp37-cp37m-manylinux1_x86_64
New WHEEL info tags: cp27-cp27m-manylinux1_i686
New WHEEL info tags: cp27-cp27mu-manylinux1_i686
New WHEEL info tags: cp36-cp36m-manylinux1_i686
New WHEEL info tags: cp37-cp37m-manylinux1_i686
```

madig:
I suppose it's thorough. Which ones for 2.7 do we not need?

anthrotype:
No, I don't think it's not too many. Those are the right ones.

.appveyor.yml
anthrotype: so you need to find out where cibuildwheel saves the built wheel package, then the CIBW_TEST_COMMAND will be `tox --installpkg $wheel`

this way tox will not install from source but will install the given wheel inside its own venv, then proceed to run the tests

anthrotype:
the windows ones are made by Appveyor. The manylinux1 wheels, well... are for Linux

anthrotype:
cibuildwheel also takes care of building wheels for macos for various pythons

madig:
@miguelsousa see e.g. https://ci.appveyor.com/project/adobe-type-tools/psautohint/build/1.0.215/artifacts

madig:
tox is more complicated... will look at it tomorrow.

anthrotype:
there's no rush

btw, appveyor say they'll ship python3.7 soon, good to know.
https://github.com/appveyor/ci/issues/2475

anthrotype:
I guess.. we could ditch the i686 (32-bit only) linux wheels... It's 2018, so it's unlikely anybody is running a 32 linux (if they I'd say they could compile from source).

miguelsousa:
the wheel building should only be triggered on tags, right?

anthrotype:
no, at every commit.
The upload only on tags.
Building wheels is not different from building from source. You have to compile it first in order to be able to run the tests.

.travis.yml
```diff
@@ -21,12 +21,13 @@ script:
   - cibuildwheel --output-dir wheelhouse
   - $PIP install -r requirements.txt
   - $PIP install -r dev-requirements.txt
-  - tox -e codecov
+  - wheel=`ls ${wheelhouse}/psautohint*.whl | head -n 1`
+  - tox --installpkg $wheel -e codecov --workdir {project} -c {project}
```
anthrotype: we don't want to run tox ourselves again, after cibuildwheel. We want the `CIBW_TEST_COMMAND` to call tox with the --installpkg option...
cibuildwheel builds the wheel, then installs it and runs the given test command. We want tox to take care of configuring the test environment, but we want to make it use a pre-build package (the wheel that we just built with cibuildwheel), instead of rebuilding the module from a sdist (which is its default behavior).
I need to go now, I'll have a look again tomorrow.

madig:
Until the test suite is trimmed to not run for 3 minutes, one could make multiple jobs for for every Python version. That might speed things up a bit.

anthrotype:
Maybe suggesting to combine tox with cibuildwheel was not a good idea.
They are kind of competing with each other for the control of the CI process.
I suggest you revert this PR to the commit before tox was added, or ditch cibuildwheel and use the alternative, more configurable toolchain, multibuild, that we use for compreffor, brotli-wheels, pyclipper, and others.
I can take over this PR if you wish.

.travis.yml
```diff
@@ -4,7 +4,7 @@ env:
   global:
     - CIBW_SKIP="cp34-* cp35-*"
     - CIBW_TEST_REQUIRES="tox"
-    - CIBW_TEST_COMMAND="tox --installpkg /tmp/built_wheel/*.whl -e codecov --workdir {project} -c {project}"
+    - CIBW_TEST_COMMAND="tox --installpkg wheelhouse/*.whl -e codecov --workdir {project} -c {project}"
```
anthrotype: You don’t want to only run `-e codecov` tox environment, neither you want to run all the envlist.. but the specific python version that is currently active. I think `-e py` (or `-e py-cov`) is the shortcut for the generic `python` executable. I’m typing from my phone..

.travis.yml
```diff
@@ -1,43 +1,29 @@
-dist: trusty
-sudo: false
-
 language: python
 
+env:
+  global:
+    - CIBW_SKIP="cp34-* cp35-*"
+    - CIBW_TEST_REQUIRES="tox"
+    - CIBW_TEST_COMMAND="tox --installpkg wheelhouse/*.whl --workdir {project} -c {project}"
```
anthrotype: The —installpkg option needs to get the full path to the single wheel. You need to evaluate the glob pattern in a subshell.. it’s getting too complicated. Probably not worth it.

miguelsousa:
going to leave this alone now. The PR is all yours

khaledhosny:
Personally I’m more than happy to only build Linux wheels for python 3.6+, anyone using older versions should either use distro packaged `psautohint` if available or build from source.

anthrotype:
going to work on an alternative PR that uses https://github.com/matthew-brett/multibuild/ to compile the wheels. I'll also try to use pytest-xdist plugin to run the test suite itself in parallel (not just running parallel CI pipeline, but within each, using the multi-core features to speed up running the test suite).

PR #43, opened 2018-06-10
[argparse] Parse options using argparse

### LINKED ISSUE #28: Use argparse
Switching to `argparse` will break backwards compatibility with many existing options, so we have the chance to rethink them all. This work should be completed before #19.

miguelsousa:
Significant changes:
- many option arguments have changed or got removed
- "quiet mode" is now the default
- the "history file" functionality was removed
- the "counter hints file" functionality was removed (the same can be achieved via fontinfo file)

Fixes #28

### DIFF
--- a/python/psautohint/__main__.py
+++ b/python/psautohint/__main__.py
@@ -1,246 +1,91 @@
 # Copyright 2014 Adobe. All rights reserved.
 
-from __future__ import print_function, absolute_import
-
-import os
-import re
-import sys
-
-from psautohint import _psautohint, autohint, ufoFont
-
-__usage__ = """
-Auto-hinting program for PostScript and OpenType/CFF fonts.
-psautohint.py v1.50.0 Feb 11 2018
-
-psautohint -h
-psautohint -u
-psautohint -hfd
-psautohint -pfd
-psautohint [-g <glyph list>] [-gf <filename>] [-xg <glyph list>]
-           [-xgf <filename>] [-cf path] [-a] [-logOnly] [-log <logFile path>]
-           [-r] [-q] [-qq] [-c] [-nf] [-ns] [-nb] [-wd] [-o <output font path>]
-           FONT
-
+"""
+Auto-hinting program for PostScript, OpenType/CFF and UFO fonts.
 """
 
-__help__ = __usage__ + """
+from __future__ import print_function, absolute_import
 
-Takes a list of fonts, and an optional list of glyphs, and hints the fonts. If
-the list of glyphs is supplied, the hinting is limited to the specified glyphs.
+import argparse
+import os
+import plistlib
+import re
+import sys
+import textwrap
 
-Note that the hinting is better if the font's global alignment zones are set
-usefully; at the very least, you should have entered values that capture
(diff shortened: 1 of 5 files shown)


python/psautohint/autohint.py
```diff
@@ -30,15 +30,18 @@
...
+import sys
 import time
-import plistlib
 import traceback
-import shutil
 
-from psautohint import psautohint
+from fontTools.ttLib import TTFont, TTLibError, getTableModule
+
+from .psautohint import autohint, autohintmm
```
khaledhosny: I personally prefer `import foo` and `foo.Bar` style  to `from .foo import Bar` as it makes it a bit clear at the calling site where `Bar` is coming from, and keeps the global namespace a bit tidier.

python/psautohint/__main__.py
```diff
@@ -386,301 +232,522 @@
...
+        parser.error("found duplicate font inputs")
+
+    if parsed_args.reference_font:
+        all_font_paths = parsed_args.font_paths + [parsed_args.reference_font]
+    else:
+        all_font_paths = parsed_args.font_paths
+
+    options = ACOptions(parsed_args)
+
+    options.font_format = _validate_font_paths(all_font_paths, parser)
```
khaledhosny: Is this used anywhere?

python/psautohint/__main__.py
```diff
@@ -1,246 +1,91 @@
...
+from psautohint._psautohint import version as PSAUTOHINT_VERSION
+from psautohint.autohint import hintFiles, ACFontError, ACHintError, logMsg
+from psautohint.ufoFont import UFOParseError
+from psautohint.ufoFont import kProcessedGlyphsLayer as PROCD_GLYPHS_LAYER
+
+__version__ = PSAUTOHINT_VERSION
+
 
-from psautohint import _psautohint, autohint, ufoFont
+SUPPORTED_FONT_FORMATS = ('PFA', 'PFB', 'CFF', 'OTF', 'UFO')
```
khaledhosny: Do we really support Type 1 fonts? AFAIK that support was dropped a while ago, so this and the help lines below are a bit misleading until support for Type 1 is back.

python/psautohint/autohint.py
miguelsousa: I prefer it this way because it makes it clear right at the top which pieces the file is using. Doing a general import at the top and only using parts of it downstream makes the imports opaque IMO.

python/psautohint/__main__.py
miguelsousa: Not yet, but definitely will be.

python/psautohint/__main__.py
miguelsousa: Not yet, but we will. `psautohint` must support Type 1 otherwise we'll need to keep `autohint` alive for a lot longer. The Type 1 support in **autohint** was done via **tx**; in `psautohint` it will be accomplished via **fontTools**. We've somewhat discussed it on #32.

miguelsousa:
> My biggest concern is in `autohint.py`'s `openOpenTypeFile()`

Don't worry, that's all due for an overhauling.

PR #49, opened 2018-07-02
Switch to tox and pytest for the tests

khaledhosny:
Switch to tox and pytest and drop the dependency on Unix command line tools.


khaledhosny:
Any idea how to silence the Codacy warnings or should I just ignore them?

miguelsousa:
Look at them and ignore when necessary.

I have one more change coming.

tox.ini
```diff
@@ -0,0 +1,36 @@
...
+[testenv:codecov]
+changedir =
+passenv = *
+deps =
+    coverage
+    codecov
+skip_install = true
+ignore_outcome = true
+commands =
+    cd tests
```
khaledhosny: Shouldn’t we set `changedir = tests` above instead?

miguelsousa: The upload to Codecov is still not working
```
codecov runtests: commands[0] | cd tests
WARNING:command failed but result from testenv is ignored
  cmd: InvocationError for command could not find executable 'cd'
codecov runtests: commands[1] | coverage combine
No data to combine
WARNING:command failed but result from testenv is ignored
  cmd: InvocationError for command '/home/travis/build/adobe-type-tools/psautohint/.tox/codecov/bin/coverage combine' (exited with code 1)
codecov runtests: commands[2] | codecov --env TOXENV
```

khaledhosny: I’m not that familiar either, but I think you can’t freely use Unix commands here you need to whitelist them somehow. Let me try my suggestion.

khaledhosny:
Codecov seems to be working now, or did I miss something?

miguelsousa:
yeah, Codecov is receiving the coverage data now. See the coverage changes of this PR at https://codecov.io/gh/adobe-type-tools/psautohint/pull/49/tree

.appveyor.yml
```diff
@@ -1,12 +1,13 @@
...
-build_script:
-  - make
+build: false
 
 test_script:
-  - make check
+  - tox
+
+on_success:
+  - tox -e codecov
```
khaledhosny: This does not seem to work, see: https://ci.appveyor.com/project/adobe-type-tools/psautohint/build/1.0.187#L193

khaledhosny:
OK, I see a slight coverage decrease because I’m not calling __main__ anymore. Hmm, I think I’ll ignore for now.

.appveyor.yml
miguelsousa: right, the command `codecov --env TOXENV` from tox.ini is failing with,

```
The filename, directory name, or volume label syntax is incorrect.
FIND: Parameter format not correct
```

I think **TOXENV** needs to be defined somewhere in the `.appveyor.yml` file.

miguelsousa:
> OK, I see a slight coverage decrease because I’m not calling main anymore. Hmm, I think I’ll ignore for now.

That's fine. We can add tests just for that file later. Those tests won't even have to depend on the test fonts.

.appveyor.yml
khaledhosny: Still no luck. Looks Like it is using Unix `find` syntax with Windows tool with the same name. Now I wonder if this is meant to run on Windows at all…

miguelsousa: The changes may need to be more elaborate than https://github.com/adobe-type-tools/psautohint/pull/49/commits/d922b52c838e84f3ad4c6c03a5dacabf7727677e .
The fonttools configuration may provide some clues on what else is needed https://github.com/fonttools/fonttools/blob/master/.appveyor.yml

miguelsousa: On a closer look, it's working on the fonttools repo either https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.1849/job/dyslxb2m2hl1vgah

And there are other people reporting the same problem,
https://github.com/codecov/codecov-node/issues/39#issuecomment-282522840
https://github.com/codecov/codecov-node/issues/2#issuecomment-287667265

miguelsousa: That message (`FIND: Parameter format not correct`) happens even after hardcoding the file to upload, but later on the report says,
```
  codecov: commands succeeded
  congratulations :)
```
Using just `codecov --env TOXENV` produces the same results, so I think you can nuke the last commit.

miguelsousa: @khaledhosny you should definitely remove my last commit https://github.com/adobe-type-tools/psautohint/pull/49/commits/c15c7c9faffe1edc6442a1a37176026964af867d because it shows up as an error at https://codecov.io/gh/adobe-type-tools/psautohint/commits
 
And [this page](https://codecov.io/gh/adobe-type-tools/psautohint/commit/6d103e8b45b8eb09d446e0f1493ef3e21792bcfd/build) confirms that Codecov received all the uploads from your last commit https://github.com/adobe-type-tools/psautohint/pull/49/commits/6d103e8b45b8eb09d446e0f1493ef3e21792bcfd

khaledhosny: OK, removed. I think we can merge now.

miguelsousa: 👍

anthrotype:
@khaledhosny I just saw you pinged me, I was away last week. Overall this looks good to me! :+1:

PR #51, opened 2018-07-03
Fix Travis branch regex

khaledhosny:
This seems to expect a regular expression not a glob pattern.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,4 @@ after_success:
 branches:
   only:
     - master
-    - wip/*
+    - /^wip\/.*$/


anthrotype:
careful when you use the `branches:` option in .travis.yml. If will prevent you from running CI builds when you push a git tag.
A git tag is interpreted by Travis as a branch with the given tag name.
You need to whitelist those, like we do here e.g.:

https://github.com/googlei18n/ufo2ft/blob/dbfe64c35592725a04faf06ce67a456827966e2a/.travis.yml#L12

cc @cjchapman knows

khaledhosny:
Thanks @anthrotype!

PR #52, opened 2018-07-03
Whitelist tags on Travis as well

khaledhosny:
Per Cosimo’s comment:
https://github.com/adobe-type-tools/psautohint/pull/51#issuecomment-402110838

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,7 @@ after_success:
 branches:
   only:
     - master
+    # We want to build wip/* branches since these are not usually used for PRs
     - /^wip\/.*$/
+    # We want to build version tags as well.
+    - /^v\d+\.\d+.*$/


miguelsousa:
The regexp you want is actually `/^\d+\.\d+\.\d+.*$/`
https://github.com/adobe-type-tools/afdko/blob/develop/.travis.yml#L91

miguelsousa:
That version is already released, so it shouldn't be a problem. Going forward all tags should abide by [PEP 440](https://www.python.org/dev/peps/pep-0440/).

anthrotype:
The git tag can be prefixed with v, that’s very common. The PEP440 is for the package version proper only. For example, setuptools_scm automatically strips the leading v when generating the package version from the git tag, because it assumes that the git tag will have a v prefix. My preference is for the v to be present in the git tag (It helps identifying among other tags, and as Khaled said we have already pushed tags like the that), but you decide.

khaledhosny:
What @anthrotype said.

miguelsousa:
I don't see the point of having the `v`, but if you both prefer that I'm fine with it. My goal was to get psautohint to adopt a similar configuration as afdko, because maintaining two different processes is onerous.

Issue #39, opened 2018-04-14
Add tests for multiple master hinting


khaledhosny:
We don’t currently run any tests for this code.

PR #53, opened 2018-07-03
Add test for compatible (MM) hinting

khaledhosny:
Fixes https://github.com/adobe-type-tools/psautohint/issues/39

### DIFF
--- a/python/psautohint/__main__.py
+++ b/python/psautohint/__main__.py
@@ -16,7 +16,8 @@ import textwrap
 from fontTools.misc.py23 import open
 
 from psautohint._psautohint import version as PSAUTOHINT_VERSION
-from psautohint.autohint import hintFiles, ACOptions, ACFontError, ACHintError, logMsg
+from psautohint.autohint import (ACFontError, ACHintError, ACOptions,
+                                 hintFiles, logMsg)
 from psautohint.ufoFont import UFOParseError
 from psautohint.ufoFont import kProcessedGlyphsLayer as PROCD_GLYPHS_LAYER
 
--- a/python/psautohint/__main__.py
+++ b/python/psautohint/__main__.py
@@ -331,7 +331,7 @@ class Options(ACOptions):
     def __init__(self, pargs):
         super(Options, self).__init__()
         self.inputPaths = pargs.font_paths
-        self.outputPath = pargs.output_path
+        self.outputPaths = pargs.output_paths
         self.reference_font = pargs.reference_font
         self.hintAll = pargs.hint_all_ufo
         self.verbose = pargs.verbose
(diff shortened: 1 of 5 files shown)


python/psautohint/__main__.py
```diff
@@ -682,7 +682,9 @@ def get_options(args):
     )
     parsed_args = parser.parse_args(args)
 
-    if len(parsed_args.font_paths) > 1 and parsed_args.output_path:
+    if (parsed_args.output_paths and
+            len(parsed_args.font_paths) != len(parsed_args.output_paths)):
+        print(parsed_args)
```
miguelsousa: is this a debug print?

python/psautohint/__main__.py
```diff
@@ -682,7 +682,9 @@ def get_options(args):
...
         parser.error("-o/--output can't be used with multiple font inputs")
```
miguelsousa: the check changed so the message should change accordingly

tests/test_hintfontsmm.py
```diff
@@ -0,0 +1,28 @@
...
+        self.outputPaths = outpaths
+        self.reference_font = reference
+        self.hintAll = True
+        self.verbose = False
+
+
+@pytest.mark.parametrize("base", glob.glob("data/*/*Masters"))
+def test_mmufo(base):
+    paths = sorted(glob.glob(base + "/*.ufo"))
+    options = Options(paths[0], paths[1:], [p + ".out" for p in paths[1:]])
```
miguelsousa: better to define `reference`, `inpaths` and `outpaths` outside, I think

miguelsousa:
please wrap the imports on this line as well https://github.com/adobe-type-tools/psautohint/pull/53/files#diff-df7531b6e02c6fd7b0cdebea307ebfb6R19

tests/test_hintfontsmm.py
khaledhosny: Naming variables is hard, I struggled with finding appropriate names so I avoided them altogether :) Using your suggestion now.

tests/test_hintfontsmm.py
```diff
@@ -19,8 +19,12 @@ def __init__(self, reference, inpaths, outpaths):
 @pytest.mark.parametrize("base", glob.glob("data/*/*Masters"))
 def test_mmufo(base):
     paths = sorted(glob.glob(base + "/*.ufo"))
-    options = Options(paths[0], paths[1:], [p + ".out" for p in paths[1:]])
+    reference = paths[0]
+    inpaths = paths[1:]
+    outpaths = [p + ".out" for p in paths[1:]]
```
miguelsousa: you can use `inpaths` here instead of `paths[1:]`

python/psautohint/__main__.py
```diff
@@ -16,7 +16,8 @@
 from fontTools.misc.py23 import open
 
 from psautohint._psautohint import version as PSAUTOHINT_VERSION
-from psautohint.autohint import hintFiles, ACOptions, ACFontError, ACHintError, logMsg
```
miguelsousa: wrap using parenthesis instead
```python
from psautohint.autohint import (hintFiles, ACOptions, ACFontError, ACHintError,
                                 logMsg)
```

khaledhosny: Done.

miguelsousa: 👍

Issue #50, opened 2018-07-02
Codacy is misconfgured?


khaledhosny:
Opening https://app.codacy.com/project/adobe-type-tools/psautohint/dashboard, I see this message:

>  Hey Seems like we stopped receiving your coverage results. Check your CI server! 

And the latest changes/pull requests are not shown. However the pull requests seems to point to https://app.codacy.com/project/frankrolf/psautohint/dashboard, which suggests a misconfiguration somewhere.

miguelsousa:
The lack of coverage results is expected, because of #36. Codecov provides a lot more insight into the coverage results across branches, commits and pulls, so there's really no need to use Codacy's coverage analysis feature.

The result of the PR is at https://app.codacy.com/app/adobe-type-tools/psautohint/pullRequest?prid=1850543 . Got that link from the review of #49.
I don't know where you got that other link from; Frank is not even a member of Codacy's adobe-type-tools organization account.

One thing that Codacy doesn't do by default is branch analysis; it has to be switched on for each branch, as you can see from the screenshot below. That page is at https://app.codacy.com/app/adobe-type-tools/psautohint/settings/branches . Let me know if you can't get to it.
<img width="1104" alt="branches" src="https://user-images.githubusercontent.com/2119742/42186314-0c8d7244-7e01-11e8-830b-8cb67a8a3953.png">

khaledhosny:
No idea how I got that link from either, I don’t get it anymore. But if everything is working as intended, then sorry for the noise.

khaledhosny:
I think I found where I got that link, look at the list of checks at the end of #49, and check the “Codacy/PR Quality Review” link, here I get https://app.codacy.com/app/frankrolf/psautohint/pullRequest?prid=1850424.

khaledhosny:
![image](https://user-images.githubusercontent.com/93914/42191113-52e84d94-7e61-11e8-95d6-84c9a7cb427a.png)

(See the link shown in the status bar).

miguelsousa:
strange. For me that same link points to https://app.codacy.com/app/adobe-type-tools/psautohint/pullRequest?prid=1850543
Going to logout from GH and see what I get.

miguelsousa:
oh, I just saw it without logging out. Let me check the repository's webhooks

miguelsousa:
Not sure what's going on. @frankrolf may need to do something on his account.
[...]

PR #55, opened 2018-07-05
setup multibuild for compiling multi-platform wheels

anthrotype:
follow up from #42


anthrotype:
codacy is very picky.. didn't know it also linted bash! :)

anthrotype:
appveyor is very slow, not so much the single build (5-6 min each) but the fact that unlike Travis doesn't run in parallel with the other matrix builds.
We might decide to drop 64-bit 2.7 and 32-bit 3.x, on the reasoning that legacy is legacy, and new OSes ==> modern python architecture?..

anyway, i'll complete this tomorrow.

anthrotype:
yes Codacy, I _do_ want to shadow the built-in `open` function, and the use of `ElementTree.parse` function is totally ok. Shut up.

anthrotype:
I'm not an admin on the appveyor account, so I cannot store my encrypted PyPI password as `TWINE_PASSWORD` secure environment variable in appveyor.yml.
For now I added the same encrypted credentials as the adobe-type-tools/afdko project. I believe that the secure encrypted variables in appveyor are the same across project (unlike Github), only vary based on the Appveyor account with which they are encrypted, and psautohint and afdko should belong to the same account.
We'll see if this work after we merge this PR and try to push a pre-release tag.

miguelsousa:
The Codacy review link is pointing to the wrong page; #50 is tracking that. The correct link is https://app.codacy.com/app/adobe-type-tools/psautohint/pullRequest?prid=1867014

I'll look into providing twine credentials for this repo.

anthrotype:
ok the pr is ready for review/merge now. I'm sure there'll be more fixes needed after we try to push the next tag.

anthrotype:
for Travis I added my PyPI username/password since I'm a collaborator on both this repo and the linked PyPI page, so they will work.
If you prefer to replace them with yours, I'm ok. Just make sure to use this command:

```sh
$ echo "<your-password-here>" | travis encrypt -R adobe-type-tools/psautohint --add deploy.password
```

Then you'll notice that the .travis.yml file will have been reformatted with indentation and empty line separators all messed up. What I usually do is copy/paste the `secure: ...` password line somewhere else, then restore the .travis.yml file to its original state before the travis tool reformatted it (`git checkout .travis.yml`), and then paste the encrypted password to its correct place.

.travis.yml
```diff
@@ -24,6 +24,8 @@ env:
     - TEST_DEPENDS="-rdev-requirements.txt"
     - PLAT=x86_64
     - UNICODE_WIDTH=32
+    - TWINE_USERNAME=adobe-type-tools-ci
+    # TWINE_PASSWORD is set in Travis settings
```
anthrotype: @miguelsousa are you sure that setting these environment variable works with Travis deploy too? 
It does also use twine under the hood, so it should work. But I've never tried. Anyway we'll see

anthrotype:
oops you merged too fast, or we didn't coordinate :)

https://ci.appveyor.com/project/adobe-type-tools/psautohint/build/1.0.246/job/dojxioahaa9nd753#L198

I added some commits that broke the appveyor build

anthrotype:
@miguelsousa I pushed the fix to master, can you please stop the previous Appveyor builds, so that let the last one can start? I can start Travis builds, but not Appveyor's

anthrotype:
the last couple of commits to .appveyor.yml were a minor optimization, to avoid having to compile the module from source twice: i.e. once when tox installs it into its test environments, and again before uploading the wheel to PyPI.
I changed it so that we build the wheel before running the tests, then call tox with `--installpkg` option, passing the full path to the compiled wheel file (with that `dir` command hack that uses a wildcard to get the filename, writes to a temp file, then reads the content into an environment variable.. windows batch can be weird, maybe a powershell command could have done the same thing, but I don't know that).
Then, when we upload with twine, we already have the wheel (which we just built and tested) so we can upload the same file without recompiling it again.

it's working now:
https://ci.appveyor.com/project/adobe-type-tools/psautohint/build/1.0.248/job/0a8hx3250c5464o2#L205

anthrotype:
@miguelsousa I'll work on adding python3.7 wheels later (appveyor doesn't ship with 3.7 yet, but we can download/install it ourselves). I think you can try tagging a prerelease and see it pushes the wheels. It _never_ works at first try -- no matter how many times we set this things up over and over :smile:

miguelsousa:
go for it

miguelsousa:
@anthrotype I'm not sure what tags you guys want to use, so you go ahead with the prerelease

anthrotype:
ok, I'll try in a bit. I noticed that we have two sets of version strings, one in setup.py (1.1.1.dev0) which is in sync with the git tag, and another one for the libpsautohint C library (1.6.0). The psautohint --version currently prints the latter.
Do we want to keep the two distinct. Or would it be better to merge them?
I'd prefer a single version for both, since the C library is not shipped separately from the python wrapper. 
We could use setuptools_scm to keep that in sync with the git tags, and then we could be able to cut a release by simply pushing a tag (without needing to manually bump the version string in multiple places).
setuptools_scm has a feature that can automatically writes the computed version string to a text file, in our case this could be a C header which the `AC_getVersion()` function can use, instead of hard-coding it as it is now.

/cc @khaledhosny

miguelsousa:
@anthrotype I prefer a single version as well. It should match the version of the wheel.

.travis.yml
anthrotype: actually, it failed to deploy without 'user' defined:
https://travis-ci.org/adobe-type-tools/psautohint/jobs/401042408#L878

I fixed it here: https://github.com/adobe-type-tools/psautohint/commit/65f9234828f75e57fb69d41091103a77ccf5daa8

:crossed_fingers:

anthrotype: .. And also the 'password' field is required:
https://travis-ci.org/adobe-type-tools/psautohint/jobs/401053332#L878

I'm using mine just to test. The reason that alternative method works is because it calls twine directly (like we do with Appveyor), whereas in Travis we are using the `dpl` tool (which requires those settings in the .travis.yml, and in turns calls twine with them).

miguelsousa: 👌
the afdko repo uses a script to deploy to PyPI from Travis
https://github.com/adobe-type-tools/afdko/blob/develop/.travis.yml#L117

anthrotype: @miguelsousa can you please cancel this appveyor build for me?
https://ci.appveyor.com/project/adobe-type-tools/psautohint/build/1.0.256

I don't seem to have access to it, even despite I saw your invite in my email. The address is correct, not sure why it isn't working :-/
I'm logged in Appveyor with my github username @anthrotype

anyway we need to stop that appveyor build before travis pushes a new release which is going to be newer than the one appveyor is trying to push :(

miguelsousa: @anthrotype cancelled. Try logging into AppVeyor using your email address. The invitation is still showing up as not having been accepted.

anthrotype: I did! The link asks me to "sign up" but of course I am already signed up... so I click the button to login with github and select "anthrotype" among several organizational accounts (e.g. fonttools, trufont, etc. but not afdko...), but then nothing happens. I don't seem to be able to cancel or restart the builds. Anyway, no problem. no big deal

miguelsousa: @anthrotype I've reconfigured the access to AppVeyor to use GH teams. Accept the invitation to https://github.com/orgs/adobe-type-tools/teams/psautohint-team and then check AppVeyor to see if you're able to restart builds now

anthrotype: Thanks Miguel! Yeah, that should do it.

btw, I'm happy to report that auto-upload to pypi from travis/appveyor is now working:
https://pypi.org/project/psautohint/1.1.1a6/#files

khaledhosny:
This seems to have squashed and merged as a single commit, something I strongly dislike myself. Though keeping each of the original commits might be too much, at least commits should be grouped into some reasonable units e.g. why opening files in UTF-8 encoding is in the same commit with the changes to the test suite? If one of these two unrelated changes proved problematic in the future, reverting that bit will be unnecessarily complicated, the same goes for git bisecting to find the source of a regression.

anthrotype:
I agree with Khaled. I never click merge&squash button on Github. I prefer to squash commits manually into logical groups. It was inevitable that there were many small commits because of the trial and error when dealing with CIs. Next time there's a big PR like this, we could let the dev do the squashing manually after the reviewer has approved.

khaledhosny:
No objection to single version, it should probably be based on the current C library version. So probably 1.7.0something.

anthrotype:
btw, the changes to open files as UTF-8 were necessary in order to let the tests pass on the manylinux1 docker which sets LANG=C locale and thus open text files as ascii by default, unless explicitly told encoding="utf-8".

miguelsousa:
Sorry, going forward I'll leave the merging part up to you two. I prefer to merge&squash when there are too many minute changes that are toward the same goal.

PR #58, opened 2018-07-07
More tests

khaledhosny:
Tests for CFF and PFA fonts (the later simply raise an exception), in addition to some cleanup.

I’m starting to move shared and cleaned up code to `psautohint/__init__.py` with the hope that it eventually replaces `psautohint/autohint.py` and provide the new module API. WDYT?

### DIFF
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -1107,13 +1107,11 @@ class CFFFontData:
                         os.remove(inputPath)
                         os.rename(tempPath, inputPath)
                     except (OSError, IOError):
-                        import traceback
-                        self.logMsg("\t%s" % (traceback.format_exception_only(
-                            sys.exc_info()[0], sys.exc_info()[1])[-1]))
                         self.logMsg(
                             "Error: could not overwrite original font file "
                             "path '%s'. Hinted font file path is '%s'." %
                             (inputPath, tempPath))
+                        raise
             else:
                 ttFont.save(outFilePath)
                 ttFont.close()
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -1,8 +1,6 @@
 # Copyright 2014 Adobe. All rights reserved.
 
 """
-otfFont.py v1.13.1 Feb 11 2018
-
 Utilities for converting between T2 charstrings and the bez data format.
 Used by AC and focus/CheckOutlines.
 """
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -1014,13 +1014,13 @@ def convertBezToT2(bezString):
 
 
 class CFFFontData:
-    def __init__(self, ttFont, inputPath, outFilePath, fontType, logMsgCB):
+    def __init__(self, ttFont, inputPath, outFilePath, font_format, logMsgCB):
         self.ttFont = ttFont
         self.inputPath = inputPath
         if (outFilePath is None):
             outFilePath = inputPath
         self.outFilePath = outFilePath
-        self.fontType = fontType
+        self.font_format = font_format
         self.logMsg = logMsgCB
         try:
             self.cffTable = ttFont["CFF "]
@@ -1089,7 +1089,7 @@ class CFFFontData:
 
     def saveChanges(self):
         ttFont = self.ttFont
-        fontType = self.fontType
+        font_format = self.font_format
         inputPath = self.inputPath
         outFilePath = self.outFilePath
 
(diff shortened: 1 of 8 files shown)


tests/test_hintfonts.py
```diff
@@ -21,9 +22,13 @@ def __init__(self, inpath, outpath):
         self.verbose = False
 
 
+def _temp(tmpdir, path):
+    return str(tmpdir / basename(path))
```
miguelsousa: this seems like an odd way of assembling a path. I'm surprised that it even works on Windows. I wonder if tox is fixing it under the hood.

tests/test_hintfonts.py
```diff
@@ -32,16 +37,41 @@ def test_ufo(ufo, tmpdir):
...
+        writer = XMLWriter(_temp(tmpdir, path) + ".xml")
+        with open(path, "rb") as fp:
+            font.decompile(fp, None)
+            font.toXML(writer)
+        del writer
+        del font
+
+    assert differ([_temp(tmpdir, cff) + ".xml", _temp(tmpdir, out) + ".xml"])
+
+@pytest.mark.parametrize("pfa", glob.glob("%s/*/*/font.pfa" % DATA_DIR))
```
miguelsousa: add a blank line above

khaledhosny:
> I'd like to see all logMsg() replaced with info/warn/debug messages from the logging module at some point.

Me too, I have it on my radar…

tests/test_hintfonts.py
khaledhosny: I don’t have the slightest idea how it works, I just reshuffled @anthrotype’s code.

anthrotype:
> Increase the version of otfFont.py and ufoFont.py, or remove them altogether.

there should be no other versions besides the package's own version (defined in setup.py). All the others are redundant.

tests/test_hintfonts.py
anthrotype: pytest `tmpdir` fixture is a `py.path.LocalPath` object, here is the api:
http://py.readthedocs.io/en/latest/path.html
https://docs.pytest.org/en/latest/tmpdir.html

anthrotype: the `py` module is part of pytest (originally, pytest was part of it, hence the old name py.test)

python/psautohint/otfFont.py
```diff
@@ -1014,13 +1012,14 @@ def convertBezToT2(bezString):
 
 
 class CFFFontData:
-    def __init__(self, ttFont, inputPath, outFilePath, fontType, logMsgCB):
+    def __init__(self, ttFont, inputPath, outFilePath, allow_decimal_coords,
+                 font_format, logMsgCB):
```
anthrotype: there  a mixture of camelCase (predominant in the original codebase) and snake_case. I guess we should stick to one style or the other.

tests/test_hintfonts.py
```diff
@@ -21,9 +22,13 @@ def __init__(self, inpath, outpath):
         self.verbose = False
 
 
+def _temp(tmpdir, path):
```
anthrotype: the name `_temp` can be misleading. that code is not making a temporary directory, is just concatenating a py.path.local object with another and returning a str. The tmpdir fixture is responsible for creating the temporary directory and cleaning it up automatically after the test function has completed. I think factoring out this one liner only obfuscates things -- well, at least for me, who am familiar with using the fixture and py.path.local.

python/psautohint/otfFont.py
khaledhosny: I’m moving away from camelCase as this code is being, slowly, rewritten.

tests/test_hintfonts.py
anthrotype: maybe `_join` or `_pjoin` is a better name

tests/test_hintfonts.py
```diff
@@ -32,16 +37,42 @@ def test_ufo(ufo, tmpdir):
...
+    hintFiles(options)
+
+    for path in (cff, out):
+        font = CFFFontSet()
+        writer = XMLWriter(_temp(tmpdir, path) + ".xml")
+        with open(path, "rb") as fp:
+            font.decompile(fp, None)
+            font.toXML(writer)
+        del writer
+        del font
```
anthrotype: is this necessary? aren't they garbage-collected automatically at each iteration, since the variables are reassigned to different objects? or are there circular references hidden somewhere?

khaledhosny: It is just an unprofiled micro optimization that I kept copying from one test to another…

khaledhosny: Hmm, dropping this breaks the tests. Something fishy is going on…

khaledhosny: OK, I need to close `XMLWriter` or written data to file will be truncated.

PR #59, opened 2018-07-07
WIP: use setuptools_scm to compute package version and write version.h

anthrotype:
there's still some work to do, as the autotools build system should also be made aware of the version.h file: it should run `python setup.py egg_info` to generate the header containing the computed version.
Will finish next week.


khaledhosny:
You can ignore the autotools build for now.

anthrotype:
ok, if we leave out the configure.ac and Makefile.am stuff, this is ready to be merged now.

if one attempts to build the old autohintexe using autotools from a fresh git checkout, it fails because "version.h" file is cannot be included as it needs to be generated first, with `python setup.py egg_info`.
From an unzipped sdist, it works because the version.h file is included in there.

We should update the instructions in README.md, or fix autotools to read the package version from version.h. I'll leave that to Khaled, as i'm not familiar enough with that build system.

anthrotype:
ugh.. the github merge button was defaulting to "rebase".. I actually prefer a merge commit for PRs. It allows you to look at the git log graph and see the history as it really unfolded.
Well, too late for this one. It seems "allow merge commit" option was disabled in the repository settings. I have re-enabled it.

anthrotype:
I've pushed v1.7.0b1 tag to verify everything is working. Then when you guys are ready for v1.7.0, you just have to push an annotated git tag (with the message containing the release notes) and boom, you're done. 
Alternatively, you can create a Github Release directly from the Github web UI, which will push a tag and trigger the CI release process.
The release git tags need to be prefixed with a "v" and then a PEP 440 compatible version string https://www.python.org/dev/peps/pep-0440/

anthrotype:
You'll notice that when creating the release from the web UI the release notes are formatted as markdown, whereas when pushing a git tag created on the console, the tag message is formatted as a block of code with monospaced font, and the tag is not automatically marked as a "Github Release" in bold, unless you click "edit tag" and manually copy-paste the git tag message into the release notes text box (which is what i've been doing for the other projects so far, though it's kind of tedious):
https://github.com/adobe-type-tools/psautohint/releases
https://github.com/adobe-type-tools/psautohint/releases/tag/v1.7.0b1

I'm considering experimenting with this script that allows to create/edit github releases from the command line via its API:

https://github.com/j0057/github-release

We could trigger it on the CI on tagged commits, so that it creates the release with the tag message (if the tag is annotated), and possibly even upload the built artifacts to github (though that's not really needed since we already have pypi).

khaledhosny:
> ugh.. the github merge button was defaulting to "rebase"

Yes, sorry for that. I have a strong dislike for merge commits that I always avoid them, to the degree that I change the repo settings to remove it so I don’t do it by mistake. But if others prefer merge commits, then OK, I can (try to) live with that.

anthrotype:
why so?
I like the merge commit for PRs especially when I need to draft the changelog for a new release, the commits are grouped by the PR branch instead of interspersed with each other in a random order.

khaledhosny:
I prefer linear git history, it is a personal preference but it feels cleaner and more straight forward, also http://www.bitsnbites.eu/a-tidy-linear-git-history/

anthrotype:
I think we can keep everyone happy if:
1) we manually squash commits into logical units (to reduce noise, and help with bisecting) just before merging a PR
2) we do a `git rebase origin/master && git push -f` locally, once a PR has been reviewed and just before merging it (this way we ensure a linear git history)
3) in the github merge button, we select the option that keeps the merge commit, which bears information as to the PR number associated with a feature branch.

khaledhosny:
That feels a bit over complicated. I’d rather go with merge commits if this what others prefer.

PR #60, opened 2018-07-08
Logging

khaledhosny:
Use logging module instead of homegrown one.

### DIFF
--- a/libpsautohint/autohintexe.c
+++ b/libpsautohint/autohintexe.c
@@ -24,6 +24,7 @@ static char* fileSuffix = NULL;
 static FILE* reportFile = NULL;
 
 static bool verbose = true; /* if true don't number of characters processed. */
+static bool debug = false;
 
 static void openReportFile(char* name, char* fSuffix);
 
@@ -110,6 +111,10 @@ static void
 reportCB(char* msg, int level)
 {
     switch (level) {
+        case -1: /* LOGDEBUG */
+            if (debug)
+                fprintf(stdout, "DEBUG: %s\n", msg);
+            break;
         case 0: /* INFO */
             if (verbose)
                 fprintf(stdout, "INFO: %s\n", msg);
@@ -229,7 +234,7 @@ main(int argc, char* argv[])
      static void Blues()
      */
 
-    bool allowEdit, roundCoords, allowHintSub, debug, badParam, allStems;
+    bool allowEdit, roundCoords, allowHintSub, badParam, allStems;
     bool argumentIsBezData = false;
     bool doMM = false;
     bool report = false;
@@ -245,7 +250,6 @@ main(int argc, char* argv[])
     int result, argi;
 
     badParam = false;
-    debug = false;
     allStems = false;
 
     allowEdit = allowHintSub = roundCoords = true;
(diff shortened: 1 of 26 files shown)


libpsautohint/autohintexe.c
```diff
@@ -402,7 +420,7 @@ main(int argc, char* argv[])
             }
 
             result = AutoColorString(bezdata, fontinfo, &output, &outputsize,
-                                     allowEdit, allowHintSub, roundCoords, debug);
+                                     allowEdit, allowHintSub, roundCoords);
```
anthrotype: Removing debug argument will break the C api of libpsautohint. I guess next version should be 2.0

libpsautohint/autohintexe.c
```diff
@@ -107,9 +108,27 @@ vstemCB(int right, int left, char* glyphName)
...
 
 static void
-reportCB(char* msg)
+reportCB(char* msg, int level)
 {
-    fprintf(stdout, "%s", msg);
+    switch (level) {
+        case -1: /* LOGDEBUG */
+            if (debug)
+                fprintf(stdout, "DEBUG: %s\n", msg);
```
anthrotype: If debug variable is constantly false, no debug logging message will be printed despite the level (unless i’m missing something)

python/psautohint/__main__.py
```diff
@@ -701,18 +709,12 @@ def get_options(args):
...
     options = get_options(args)
 
-    if options.logFile:
-        open(options.logFile, 'w', encoding="utf-8")
-
     try:
         hintFiles(options)
     except Exception as err:
-        logMsg("\t%s" % err)
+        logging.exception("%s", err)
```
anthrotype: Logging.exception already prints the traceback including the last exception error. The logged message should be something else, such as “Unhandled exception occurred”

libpsautohint/autohintexe.c
khaledhosny: We broke it many times before, and will probably do again. But since we don’t publish the C library independently I think it should be OK. One day we should stabilize it and make such a standalone release, but I want to rewrite the API before that.

libpsautohint/autohintexe.c
khaledhosny: Old code (not shown in the diff) sets it. it used to be a local variable inside `main()` but it is global now.

libpsautohint/autohintexe.c
anthrotype: sgtm

python/psautohint/__main__.py
```diff
@@ -648,6 +647,15 @@ def get_options(args):
     )
     parsed_args = parser.parse_args(args)
 
+    log_level = logging.WARNING
```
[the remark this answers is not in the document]
miguelsousa: I think ERROR messages should be seen regardless of the verbose level. I tried to test it but ran into other unrelated issues. Leave the setup of ERROR as is for now until it can be how it's working.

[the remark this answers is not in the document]
khaledhosny: OK.

khaledhosny:
@miguelsousa Can this be merged now?

miguelsousa:
@khaledhosny yes

PR #62, opened 2018-07-09
CLI tests

khaledhosny:
Fix calling command line tool with no output path, and add some CLI tests.

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -214,8 +214,11 @@ def openFile(path, outFilePath, options):
 
     if font_format == "UFO":
         font = openUFOFile(path, outFilePath, options)
-    else:
+    elif font_format in ("OTF", "CFF"):
         font = openOpenTypeFile(path, outFilePath, font_format, options)
+    else:
+        raise NotImplementedError("%s format is not supported yet, sorry." %
+                                  font_format)
 
     return font
 
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -282,10 +282,9 @@ def hintFiles(options):
     if options.reference_font:
         hintFile(options, options.reference_font, None, reference_master=True)
     for i, path in enumerate(options.inputPaths):
-        try:
+        outpath = None
+        if options.outputPaths is not None and i < len(options.outputPaths):
             outpath = options.outputPaths[i]
-        except IndexError:
-            outpath = None
         hintFile(options, path, outpath, reference_master=False)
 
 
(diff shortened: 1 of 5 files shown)


python/psautohint/autohint.py
```diff
@@ -282,10 +282,9 @@ def hintFiles(options):
     if options.reference_font:
         hintFile(options, options.reference_font, None, reference_master=True)
     for i, path in enumerate(options.inputPaths):
-        try:
+        outpath = None
+        if options.outputPaths is not None and i < len(options.outputPaths):
```
anthrotype: maybe you could add be a specific test with multiple -o options when there are multiple input_paths/output_paths
https://github.com/adobe-type-tools/psautohint/blob/6293d470ed157e018f4a067407e888977619f17e/python/psautohint/__main__.py#L489

tests/test_cli.py
```diff
@@ -34,6 +34,37 @@ def test_outpath(path, tmpdir):
...
+    referenceSrc.copy(referenceDst)
+    reference = str(referenceDst)
+    inpaths = paths[1:]
+    outpaths = [str(tmpdir / basename(p)) for p in inpaths]
+
+    psautohint(inpaths + ['-o'] + outpaths + ['-r', reference])
+
+
+def test_outpath_multi_unequal(tmpdir):
+    """Test that we exit if output paths don't match number if input paths."""
```
anthrotype: typo: "number _of_ input paths"

python/psautohint/autohint.py
```diff
@@ -214,8 +215,11 @@ def openFile(path, outFilePath, options):
 
     if font_format == "UFO":
         font = openUFOFile(path, outFilePath, options)
-    else:
+    elif font_format in ("OTF", "CFF"):
         font = openOpenTypeFile(path, outFilePath, font_format, options)
+    else:
+        log.error("%s format is not supported yet, sorry.", font_format)
+        sys.exit(1)
```
khaledhosny: I’m not sure if it is a good idea to call `sys.exit()` from the module, this would break other applications (an imaginary future version of `ufo2ft`, for example) importing and calling this module.

anthrotype: Yes, SystemExit exceptions only from main() please

miguelsousa: This is just a temporary solution. Feel free to change it. I can't think of a better way to do it right down. My only request is for it to terminate gracefully from the point of view of a user running psautohint on the command line.

khaledhosny: We can raise a specific exception, say `ACFontNotSupported` and catch that in main and fail gracefully.

miguelsousa: sounds good. I won't be able to do it until tomorrow.

khaledhosny: Or, since this is temporary until Type 1 support is restored, just `NotImplementedError`.

PR #64, opened 2018-07-11
Misc. cleanups

khaledhosny:


python/psautohint/fdTools.py
```diff
@@ -195,16 +199,16 @@ def buildBlueLists(self):
...
-                        print("Error in FDDict %s. The top of zone %s at %s "
-                              "is within the min spearation limit (%s units) "
-                              "of zone %s with the bottom at %s." %
-                              (self.DictName, prevPair[2], prevPair[0],
-                               zoneBuffer, pair[2], pair[1]))
+                        log.error("In FDDict %s. The top of zone %s at %s is "
+                                  "within the min spearation limit (%s units) "
+                                  "of zone %s with the bottom at %s.",
+                                  self.DictName, prevPair[2], prevPair[0],
+                                  zoneBuffer, pair[2], pair[1])
```
anthrotype: all these "errors".. it doesn't look like we are handling them in any sort of way, besides printing the message.
it they are something that the library can continue with maybe their severity is "warning"? Just wondering really.

python/psautohint/otfFont.py
```diff
@@ -13,7 +13,7 @@
 import sys
 
 from fontTools.misc.psCharStrings import T2OutlineExtractor, SimpleT2Decompiler
-from fontTools.misc.py23 import *
+from fontTools.misc.py23 import bytechr, byteord, open
```
anthrotype: good, I also don't like star imports very much

anthrotype:
one comment, maybe not related this specific PR, but since we are here. The test_cli.py takes too long, since it's running on all the fonts in the tests/data (which are huge and too many), but really only testing very specific things, which can be achieved by using some dummy test font or subsetted one.

anthrotype:
and why is Appveyor running twice on the PR? I thought setting `skip_branch_with_pr: true` would do the trick?

https://github.com/adobe-type-tools/psautohint/blob/7549c8485dfca4d01e55603955a9e2f5a0f2bf37/.appveyor.yml#L49

cc @miguelsousa

python/psautohint/fdTools.py
khaledhosny: Good point, I just followed “Error” in the old blindly, but warnings might make more sense indeed.

khaledhosny:
> The test_cli.py takes too long, since it's running on all the fonts in the tests/data

Does it? It just picks the first `.otf` and the first `.ufo` and runs on them.

khaledhosny:
> and why is Appveyor running twice on the PR? I thought setting `skip_branch_with_pr: true` would do the trick?

I think it doesn’t work because when I push the branch Appveyor will start running on it, then I open the PR and it runs on that one too. Later pushes to the branch only trigger the PR build. May be we should just do like we do with Travis.

anthrotype:
>> The test_cli.py takes too long, since it's running on all the fonts in the tests/data

> Does it? It just picks the first .otf and the first .ufo and runs on them.

it takes 23 seconds on my macbook pro with fast ssd drive. Bear in mind that we are using the tmpdir fixture and copying hundreds of small files for each test. If we use a smaller subset I think these tests could run in a bunch of seconds without losing coverage.

khaledhosny:
OK, will look into using smaller fonts for the CLI tests in the next PR (where I should be adding tests of CID fonts).

python/psautohint/otfFont.py
```diff
@@ -1173,15 +1168,14 @@ def getFontInfo(self, fontPSName, inputPath, allow_no_blues, noFlex,
                     blueValues = inactiveAlignmentValues
                     numBlueValues = len(blueValues)
                 else:
-                    raise ACFontError(
-                        "Error: font must have at least four values in it's "
-                        "BlueValues array for AC to work!")
+                    raise ACFontError("Font must have at least four values in "
+                                      "its BlueValues array for AC to work!")
```
miguelsousa: AC → psautohint

python/psautohint/fdTools.py
```diff
@@ -127,14 +131,14 @@ def getFontInfo(self):
 
     def buildBlueLists(self):
         if (self.BaselineOvershoot is None):
-            print("Error: FDDict definition %s is missing the BaselineYCoord/"
-                  "BaselineOvershoot values. These are required." %
-                  self.DictName)
+            log.warning("FDDict definition %s is missing the BaselineYCoord/"
```
miguelsousa: All the print("Error") in this file that got changed to log.warning() should be changed to raise FontInfoParseError

miguelsousa:
> If we use a smaller subset I think these tests could run in a bunch of seconds without losing coverage.

By processing a lot of fonts I think Khaled's goal was to have some assurance that the changes he is/was making in the C code were yielding the same results. It's easy to drop a lot of test files and check that the coverage of the Py code doesn't change; but how do we guarantee that for the C code?

miguelsousa:
> I should be adding tests of CID fonts

I made this CID version of SourceCodePro-Regular last night. makeotf's '-cn' option isn't working, so I had to jump thru a lot of hoops that involved modifying the font's source files and using mergeFonts, tx and ttx.
[SourceCodePro-Regular_CID.otf.zip](https://github.com/adobe-type-tools/psautohint/files/2186171/SourceCodePro-Regular_CID.otf.zip)

python/psautohint/fdTools.py
miguelsousa: I think this part of the code is parsing FDDict definitions that would be provided via a fontinfo file. It's better to raise an error, so much so that the actual message says _These are required_.

python/psautohint/otfFont.py
```diff
@@ -336,7 +335,7 @@ def optimizeT2Program(t2List):
     # entry is a list of [argList, opToken].
     # Matches logic in tx, and Adobe low level library.
     # Note that I am expecting only rlineto, vlineto, hlineto, vhcurveto,
-    # hvcurveto, rrcurveto from AC.
+    # hvcurveto, rrcurveto from psutohint.
```
miguelsousa: typo

khaledhosny: Fixed.

python/psautohint/otfFont.py
```diff
@@ -1068,45 +1068,13 @@ def updateFromBez(self, bezData, glyphName, width):
                         glyphName)
 
     def saveChanges(self):
-        ttFont = self.ttFont
-        font_format = self.font_format
-        inputPath = self.inputPath
-        outFilePath = self.outFilePath
-
-        overwriteOriginal = False
-        if inputPath == outFilePath:
```
miguelsousa: Can this check be really removed?

python/psautohint/otfFont.py
```diff
@@ -1070,49 +1065,16 @@ def updateFromBez(self, bezData, glyphName, width):
...
-                    with open(inputPath, "wb") as tf:
-                        tf.write(data)
-                else:
-                    with open(outFilePath, "wb") as tf:
-                        tf.write(data)
-
-            if os.path.exists(tempPath):
-                os.remove(tempPath)
+        if self.font_format == "OTF":
+            self.ttFont.save(self.outFilePath)
```
miguelsousa: if the `-o` option was NOT used, the output file path will be None, I think

python/psautohint/otfFont.py
khaledhosny: Why not? We no longer care whether we are saving to a new file or overwriting existing one.

miguelsousa: From the POV of a user we care, that's why I asked. But you say below that __init__ takes care of it

PR #68, opened 2018-07-14
Calculate coverage for C code as well

khaledhosny:
Add a tox environment, coverage-c, to be run on CI. A bit messy but seems to work and ~~not~~ now Codecov shows combined coverage for both C and Python code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,13 @@ matrix:
     - os: linux
       env:
         - MB_PYTHON_VERSION=3.7
+    - os: linux
+      before_install: false
+      install:
+        - pip install tox
+      script:
+        - tox -e coverage-c
+      after_success: false
     - os: osx
       language: generic
       env:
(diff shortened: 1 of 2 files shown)


miguelsousa:
@cjchapman it'd be great if we could enable this on the afdko repo as well. See https://codecov.io/gh/adobe-type-tools/psautohint/pull/68/tree/libpsautohint/src

khaledhosny:
AFDKO’s build scripts does not respect `CFLAGS` environment variable, so they will need a bit of refactoring before being able to support something like this.

anthrotype:
I'm curious to understand how this actually works under the hood, it appears to work like magic.
I only see those .gcda and .gcdo files being generated next to the .o object files upon building the extension module, but then how does python or pytest know what lines get covered? and what file is codecov actually uploading? :thinking: 
Another thing I would like to know is if this `--coverage` CFLAG adds any overhead to the compiled module, and whether it is advisable or not to keep it on for the released binary. We are currently running a separate tox environment for the c coverage, so it doesn't affect the wheel we upload to PyPI.

khaledhosny:
See https://gcc.gnu.org/onlinedocs/gcc/Gcov-Data-Files.html (`--coverage` option is a short-hand for the various compiler options listed in that page).

khaledhosny:
I think `--coverage` adds runtime overhead, unfortunately.

PR #69, opened 2018-07-15
tox: run codecov in separate 'coverage-codecov-c' env

anthrotype:
This allows to run 'tox -e coverage-c' locally on a dev's machine, without running codecov at the end (which fails unless run from the CI).


anthrotype:
it's working 
https://travis-ci.org/adobe-type-tools/psautohint/jobs/404162713

PR #70, opened 2018-07-17
WIP: CID tests

khaledhosny:
Add a CID-keyed test font and make CLI tests a bit faster.


miguelsousa:
Here's a version of `CID/font.otf` with the CFF table subroutinized: [font.otf.zip](https://github.com/adobe-type-tools/psautohint/files/2204455/font.otf.zip).
This was achieved with these commands:
```
tx -cff +S +b -std font.otf CFF_tb
sfntedit -a CFF=CFF_tb font.otf
```

And here's the corresponding CID Font Resource: [font.ps.zip](https://github.com/adobe-type-tools/psautohint/files/2204469/font.ps.zip).
This kind of file relates to a CID-keyed OTF, the same way as a PFA relates to a name-keyed OTF.
It was created using this command:
```
tx -t1 font.otf font.ps
```
Support for this format needs to be added to `get_font_format()`. The return value should be `PFC`, perhaps.

khaledhosny:
> Here's a version of CID/font.otf with the CFF table subroutinized

We used de-subroutinized files as this is what psautohint produces (and we check that output is identical).

> And here's the corresponding CID Font Resource: font.ps.zip

Thanks, added detection code and a test.

PR #73, opened 2018-07-30
Unit tests

khaledhosny:
Hinted and hinted fonts/glyph data are kept booth Bez and OTF formats, and we test the conversion to/from Bez as well as the hinting.


util/dump-font.py
```diff
@@ -0,0 +1,80 @@
...
+        glyph_map.append((name, glyph_path))
+
+    if not args.no_metainfo:
+        with open(os.path.join(outpath, "metainfo"), "w") as fp:
+            for name, path in glyph_map:
+                fp.write("%s\t%s\n" % (name, path))
+
+
+
+if __name__ == "__main__":
```
miguelsousa: remove one blank line above

miguelsousa:
Also, can UFOs be used instead of OTFs (or have UFOs alongside the OTFs)? I'm asking because as it stands it's not easy to modify (add/remove/adjust glyphs) the OTFs.

khaledhosny:
We don’t currently have code to extract hinting from UFOs and always get unhinted Bez data, but may be we can skip that step for UFOs.

miguelsousa:
`fontinfo`: OK, then please add line breaks to it.
`metainfo`: I don't think you'll need this for now, if we stick with `a-z0-9_` for the glyphs' names

miguelsousa:
> but may be we can skip that step for UFOs

yes, do that. I think it's important to have clear process for modifying the OTFs used in the unit tests, and starting testing the UFO codepath is not a bad idea either.

khaledhosny:
PTAL

python/psautohint/fdTools.py
```diff
@@ -114,15 +117,15 @@ def getFontInfo(self):
         keys = dir(self)
```
miguelsousa: Iterate using `for key, val in vars(self).items():` instead, and adjust the rest of the method accordingly.
I've seen `dir()` used in several other places in the afdko, and it's clearly not a good way to get to the class' properties.
This will also allow us to not have to `import types`.

python/psautohint/fdTools.py
```diff
@@ -220,14 +217,10 @@ def buildBlueLists(self):
 
     def __repr__(self):
         printStr = []
-        keys = dir(self)
-        for key in keys:
-            val = getattr(self, key)
-            if (val is None) or (
-               isinstance(val, types.MethodType)) or key.startswith("_"):
+        for key, val in vars(self).items():
+            if val is None or key.startswith("_"):
```
miguelsousa: I don't think there's any key that starts with `_` anymore.

It seems that `fdTools.py` is a trimmed down version of `ConvertFontToCID.py`. I've been working on that file to fix https://github.com/adobe-type-tools/afdko/issues/470, and this is how I rewrote `__repr__`:

```python
    def __repr__(self):
        fddict = {}
        for key, val in vars(self).items():
            if val is None:
                continue
            fddict[key] = val
        return "<%s '%s' %s>" % (
            self.__class__.__name__, fddict.get('DictName', 'no name'), fddict)
```

miguelsousa:
Are the failures expected, or they're happening because this is WIP?

python/psautohint/fdTools.py
khaledhosny: It is a stripped down version of that file indeed. BTW, I have been thinking, since  AFDKO is on PyPI now, should we drop all three copied and stripped down modules and import AFDKO’s directly (and possibly upstreaming any fixes I did)? Or is it too early for that?

khaledhosny:
A last minute change broke the tests, hopefully fixed now.

python/psautohint/fdTools.py
miguelsousa: no, if anything, `afdko` should depend on `psautohint`, not the other way around.

I plan to make `psautohint` an `afdko` dependency sometime soon. Once we're confident that `autohint` can be retired, I'll look into stripping the `afdko` parts that are no longer needed.

PR #80, opened 2018-08-14
More tests and cleanups

khaledhosny:
A few more tests and bug fixes/cleanups uncovered by the tests. They uncovered quite a few bugs actually.


tests/integration/test_cli.py
```diff
@@ -105,7 +109,7 @@ def test_glyph_list(glyphs, tmpdir):
     path = "%s/dummy/font.ufo" % DATA_DIR
```
miguelsousa: add one space before each of the 3 `#` above

tests/integration/test_cli.py
```diff
@@ -142,22 +146,22 @@ def test_missing_glyph_list(glyphs, tmpdir):
...
 def test_missing_cff_table(tmpdir):
     path = "%s/dummy/nocff.otf" % DATA_DIR
     out = str(tmpdir / basename(path)) + ".out"
 
-    with pytest.raises(ACFontError):
-        psautohint([path, '-o', out])
+    with pytest.raises(FontParseError):
+        autohint([path, '-o', out])
 
 @pytest.mark.parametrize("option,argument", [
```
miguelsousa: add one blank line above line 166

tests/unittests/test_extension.py
```diff
@@ -0,0 +1,114 @@
...
+def test_autohint_good_args():
+    _psautohint.autohint(INFO, GLYPH)
+
+
+def test_autohintmm_good_args():
+    _psautohint.autohintmm(INFO, (GLYPH, GLYPH), (NAME, NAME))
+
+
+@pytest.mark.parametrize("args", [
+  [],                          # no arguments
```
miguelsousa: this and many other lines below the `@pytest.mark` blocks in this file are indented with only 2 spaces instead of 4

tests/integration/test_hint.py
```diff
@@ -139,3 +139,35 @@ def test_seac_op(tmpdir, caplog):
...
+
+
+@pytest.mark.parametrize("path",
+                         glob.glob("%s/dummy/bad_privatedict_*" % DATA_DIR))
+def test_bad_privatedict_accept(path, tmpdir):
+    out = str(tmpdir / basename(path)) + ".out"
+    options = Options(path, out)
+    options.allow_no_blues = True
+
+    hintFiles(options)
```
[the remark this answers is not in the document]
khaledhosny: No, the font passes because of the `allow_no_blues` option here.

khaledhosny:
I’m keeping these for now:
```
[libpsautohint/src/gen.c:266]: (style) Condition 'doboth' is always false
[libpsautohint/src/gen.c:335]: (style) Condition 'doboth' is always false
[libpsautohint/src/psautohint.c:255]: (style) The function 'AC_initCallGlobals' is never used.
```

The 1st two need investigation, as it looks like they should be used. The later is part of *public* API and might be useful, need to check why we don’t use it.

tests/unittests/test_extension.py
khaledhosny: That is because some lines had too long comments, I’m rewording them now to be shorter.

python/psautohint/ufoFont.py
```diff
@@ -493,12 +508,12 @@ def readHashMap(self):
         self.hashMap = newMap
         return
 
-    def writeHashMap(self):
+    def writeHashMap(self, out_path):
         hashMap = self.hashMap
         if len(hashMap) == 0:
```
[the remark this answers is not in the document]
khaledhosny: I think I’m going to rework the `hashMap` code in the next round.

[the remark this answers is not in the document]
miguelsousa: sounds good. The logic in that code is known to be flawed. There are several checkoutlinesufo issues still to be resolved in the afdko repo that are caused by the hasMap process

PR #84, opened 2018-08-23
Test for reading hints

khaledhosny:
Test reading hints more extensively, and fix a bug uncovered by the tests.

### DIFF
--- a/python/psautohint/ufoFont.py
+++ b/python/psautohint/ufoFont.py
@@ -424,9 +424,9 @@ class UFOFontData:
     def isCID():
         return False
 
-    def convertToBez(self, glyphName, removeHints, doAll=False):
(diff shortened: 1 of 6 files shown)


python/psautohint/ufoFont.py
```diff
@@ -424,9 +424,9 @@ def getPSName(self):
     def isCID():
         return False
 
-    def convertToBez(self, glyphName, removeHints, doAll=False):
-        # convertGLIFToBez does not yet support hints;
-        # no need for removeHints arg.
+    def convertToBez(self, glyphName, read_hints, doAll=False):
+        # convertGLIFToBez does not yet support readinng hints, so read_hints
```
miguelsousa: readinng → reading

PR #85, opened 2018-08-23
Small Things

schriftgestalt:


khaledhosny:
We probably need to revise bez format spec even more, many of these operators are no longer supported.

PR #89, opened 2018-08-24
Cleanups and tests

khaledhosny:


python/psautohint/otfFont.py
```diff
@@ -341,11 +341,11 @@ def makeHintList(hints, needHintMasks, isH):
             continue
         pos1 = hint[0]
         pos = pos1 - lastPos
-        if isinstance(pos, float) and (int(pos) == pos):
+        if int(pos) == pos:
```
miguelsousa: would this be a better way to do it?
```python
if not pos % 1:
```
or if you prefer to make it more obvious,
```python
if pos % 1 == 0:
```

khaledhosny: I considered this, but it did not feel much readable than what you have now, but I can change it if yiu think it is better.

miguelsousa: I think changing it to one of the above would be better.

khaledhosny: Done.

PR #92, opened 2018-08-26
 Use ufoLib to write modified glyphs 

khaledhosny:
No more direct fiddling with XML and file system!


khaledhosny:
This causes a lost of changes to the test file due to different formatting. Locally I use ufonormalizer to normalize the input and output fonts before diffing and not changes were found, but it was too slow to use by default so I just updated all files to the new expected output.

PR #94, opened 2018-08-27
Add autohintexe to wheels

khaledhosny:
Switch to meson for building `autohintexe` and add it to the wheel. Works locally but not on CI bots, and I’m completely lost after several tries.


miguelsousa:
@khaledhosny I can't really help much with this. I've added @cjchapman but it will be a few weeks before he can look into it.

.appveyor.yml
```diff
@@ -80,8 +80,11 @@ install:
   # install tox
   - python -m pip install -r dev-requirements.txt
 
-  # install meson and ninja for autohintexe
-  - call %PYTHON3%\python.exe -m pip install meson ninja
+  # install meson and ninja for autohintexe.
+  # we install meson from master because it includes a 'meson.exe' entry point
+  # that makes it easier to run meson on Windows when installed via pip:
+  # https://github.com/mesonbuild/meson/pull/4004
+  - %PYTHON3%\python.exe -m pip install git+https://github.com/mesonbuild/meson.git#egg=meson ninja
```
khaledhosny: This should be `- call %PYTHON3%\python.exe …`.

anthrotype: i put it back, it's weird though

config.sh
```diff
@@ -6,7 +6,17 @@
 function pre_build {
     # Any stuff that you need to do before you start building the wheels
     # Runs in the root directory of this repository.
-    :
+
+    # meson requires python3, append it to the PATH
+    if $(python -c "import sys; sys.exit(not (sys.version_info[:2] < (3, 4)))"); then
+        if [ -n "$IS_OSX" ]; then
+            get_macpython_environment 3.6 venv3
```
anthrotype: actually this was not a good idea, it messes up the macos py27 environment. I'll try to revert to homebrew..

khaledhosny:
meson is still not seeing the correct compiler: `Native C compiler: cl (msvc 15.00.30729.01)` that is VS 2008, not the 2015 version we are trying to activate.

anthrotype:
yeah, it's distutils (or setuptools) messing up `os.environ`..

anthrotype:
however, one may argue that this is a feature. I mean, we are able to build the libpsautohint library with such an old compiler, can't we make the autohintexe.c also compile with VS2008?

anthrotype:
it kind of makes sense that we use the same compiler toolchain that was used to build the libpsautohint static library that's linked in the psautohint extension module. I understand this is not necessary, as the autohinthexe is standalone and doesn't link with python library (so can use a different version of the MSVC runtime library).

khaledhosny:
Makes sense, let me see if I can fix this.

khaledhosny:
Super weird, meson should be trying `cl.exe` first before other compilers. That suggests that it did not find `cl.exe` at all, or some other thing is going on. We need someone on Windows to check if  `authintexe` in these wheels actually work.

khaledhosny:
Thanks @anthrotype for the rescue! I squashed a few commits together. Feel free to do the same and merge when finished.

anthrotype:
OK, I also just squashed a few things. I'll try later to see if these autohintexe we built actually works on my local VM.

anthrotype:
the autohintexe.exe from the 3.6 and 3.7 wheels seems to be working fine. The one from the 2.7 wheel was throwing a "cannot find MSVCR90.DLL" error, even after I tried to install the "vcredist_x86.exe" from https://www.microsoft.com/en-us/download/details.aspx?id=5582

The build files include a "autohintexe.exe.manifest" which is generated by the link.exe and points to the correct MSVC runtime DLL. This needs to be either in the same directory or embedded inside the executable (as explained here https://msdn.microsoft.com/en-us/library/ms235591.aspx). For now let's just copy it over.

PR #97, opened 2018-09-03
setup.py: use setuptools instead of meson to build autothintexe

anthrotype:
Adding meson as a required build requirement to install psautohint from source is too heavy.

Meson requires python3 so it cannot be specified like the other python build requirements (e.g. setuptools-scm) in the pyproject.toml, because we still care about supporting python2.7 for a while, and meson can't be pip-installed on 2.7.

This means that the user who wishes (or needs, because we only provide some wheels) needs to jump through hoops to installs meson separately and ensure it's on the PATH. On windows one also needs to run inside the correct Visual Studio command prompt so that cl and the rest are in the PATH.

Now, if we are happy to compile the libpsautohint C static library using setuptools only, and use it with the psautohint extension module, then we should be fine with using the Distutils' `CCompiler.link_executable` to create autohintexe executable.

Also, the latter needs not be installed in the Python's bin or Scripts, because (as far as I know) it is not a user-facing script (unlike, say, tx), but can only be used through the old afdko's autohint.py wrapper.

There's no need to hack distutils build_scripts and install autohintexe alongside other text-only 'scripts'.
Installing binary executables as "scripts" breaks too many assumptions in the python packaging ecosystem. 

We can distribute autohintexe inside the psautohint package (just like a normal extension module) and clients who care about running it as subprocess can get its filename relative to `psautohint.__file__` (we are not zip_safe anyway).
If we do care about these being in the path we could have a custom install command that creates a symlink or a .bat wrapper.

### DIFF
--- a/python/psautohint/__init__.py
+++ b/python/psautohint/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, absolute_import
 
 import os
+import sys
 import plistlib
 
 from fontTools.misc.py23 import tobytes, tounicode
@@ -69,3 +70,14 @@ def hint_compatible_bez_glyphs(info, glyphs, masters):
                                     tuple(tobytes(m) for m in masters))
 
     return [tounicode(g) for g in hinted]
+
+
+AUTOHINTEXE = os.path.join(
+    os.path.dirname(__file__),
+    "autohintexe" + (".exe" if sys.platform in ("win32", "cygwin") else "")
+)
+if not os.path.isfile(AUTOHINTEXE) or not os.access(AUTOHINTEXE, os.X_OK):
+    import warnings
+    warnings.warn(
+        "embedded 'autohintexe' executable not found: %r" % AUTOHINTEXE
+    )
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,7 +5,6 @@ environment:
   global:
     TWINE_USERNAME: "adobe-type-tools-ci"
     # TWINE_PASSWORD is set in Appveyor settings
-    PYTHON3: "C:\\Python36-x64"
 
   matrix:
     - PYTHON: "C:\\Python27"
@@ -64,7 +63,7 @@ install:
 
   # install required python version if not present, and prepend to the PATH
   - ps: multibuild\install_python.ps1
-  - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%;%PYTHON3%\Scripts
+  - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
 
   # check that we have the expected version and architecture for Python
   - python --version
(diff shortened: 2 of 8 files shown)


khaledhosny:
I didn’t know one can do that with distutils! However, I don’t feel easy about how much code this adds to `setup.py`, but not strong enough. Please merge when ready.

khaledhosny:
Also I think we *do* need `autohintexe` to be on the `PATH` otherwise AFDKO won’t be able to use it.

anthrotype:
well, it's only a small change to have the afdko run the subprocess with the full path to the embdedded autohintexe inside psautohint package, as opposed to expecting it to be on the PATH...

anthrotype:
> I don’t feel easy about how much code this adds to setup.py

I know, it's big. In theory we could move it to a separate package (e.g. `setuptools-build-exe` or similar) and require that latter as build requirement. But life is too short. We could alternatively separate into its own module, in the same source tree (e.g. we create a `setup` folder containing a `build_exe.py`) from which setup.py imports from.

anthrotype:
another advantage of this approach over the meson one is that with here we only build the libpsautohint static library once, and link the same file into both the extension module and the executable. Whereas in the other solution, we were building it twice (with potentially different build configuration and other discrepancies).

anthrotype:
@khaledhosny do you have any idea why we get linker errors (about libm) but only when running with CFLAGS="--coverage"?
https://github.com/adobe-type-tools/psautohint/pull/97/commits/2359bf63bba3161fd9679a1c760d5562db600d6f

khaledhosny:
No idea, but I think `-lm` needs to be after `-lpsautohint` when we are statically linking (at least with some linkers).

anthrotype:
it takes ~45 minutes for appveyor to complete, I guess I'll keep the typo "AUTOHINTHEXE" in the commit message :grin:

anthrotype:
Hm, it looks like running tests on two processes is a bit faster in the end: not much, e.g. the py27 Appveyor build slowed down from 14 (2 procs) to 17 minutes (1 process).
I'm going to revert the last commit and leave the num process as is (defaults to "auto").

I'm also going to squash and merge when the build completes.

I still think that it's a bit overkill that the entire CI run takes an hour to complete, but anyway.

anthrotype:
a thin subprocess wrapper around the embedded autohintexe is now installed in the bin or Scripts folder, so it's in the PATH and can be run directly from the console.
https://github.com/adobe-type-tools/psautohint/pull/97/commits/997b2522e0e5cd5e86f88169c3159ab5c5e11b17

python/psautohint/__init__.py
```diff
@@ -69,3 +70,14 @@ def hint_compatible_bez_glyphs(info, glyphs, masters):
...
+
+AUTOHINTEXE = os.path.join(
+    os.path.dirname(__file__),
+    "autohintexe" + (".exe" if sys.platform in ("win32", "cygwin") else "")
+)
+if not os.path.isfile(AUTOHINTEXE) or not os.access(AUTOHINTEXE, os.X_OK):
+    import warnings
+    warnings.warn(
+        "embedded 'autohintexe' executable not found: %r" % AUTOHINTEXE
+    )
```
khaledhosny: Lets not bother with this here and just provide the binary. AFDKO’s `autohint.py` can have all this magic.

anthrotype: Keepin it doesn’t hurt though

anthrotype:
Anyway we can remove it later if you want.

PR #100, opened 2018-09-04
Support Type 1

### LINKED ISSUE #81: Support Type 1 font formats
[fontTools can open PFAs](https://github.com/fonttools/fonttools/blob/master/NEWS.rst#22-released-2008-05-18), and judging from the [t1Lib tests](https://github.com/fonttools/fonttools/blob/master/Tests/t1Lib/t1Lib_test.py) it can open PFBs as well, so I expect it should be possible for `psautohint` to support Type 1 font formats via fontTools.

In `autohint` the support was implemented by making a call to **tx** to convert the Type 1 font to CFF:
```
tx -cff +b -std font.pfa font.cff
```

khaledhosny:
Support Type 1 fonts using tx, plus some small cleanups.

I kept the failed attempt at using FonTools in the commit history for future reference, just in case.

Fixes #81.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -6,7 +6,9 @@
 function pre_build {
     # Any stuff that you need to do before you start building the wheels
     # Runs in the root directory of this repository.
-    :
+    if [ \( -n "$IS_OSX" -a "$MB_PYTHON_VERSION" = "2.7" \) -o \( "$PYTHON_VERSION" = "3.6" \) ]; then
+        pip install --pre afdko
+    fi
 }
 
 function run_tests {
--- a/config.sh
+++ b/config.sh
@@ -6,9 +6,12 @@
 function pre_build {
     # Any stuff that you need to do before you start building the wheels
     # Runs in the root directory of this repository.
-    if [ \( -n "$IS_OSX" -a "$MB_PYTHON_VERSION" = "2.7" \) -o \( "$PYTHON_VERSION" = "3.6" \) ]; then
-        pip install --pre afdko
-    fi
+
+    # Travis only clones the latest 50 commits. We need the full repository
+    # to compute the version string from the git metadata:
+    # https://github.com/travis-ci/travis-ci/issues/3412#issuecomment-83993903
+    # https://github.com/pypa/setuptools_scm/issues/93
+    git fetch --unshallow
 }
 
 function run_tests {
@@ -26,6 +29,11 @@ function run_tests {
     # select tox environment based on the current python version
     # E.g.: '2.7' -> 'py27-cov'
     TOXENV="py${MB_PYTHON_VERSION//\./}-cov"
+    # append the "-tx" tox factor to run the tests with afdko, but only
+    # for python 2.7 (at least for the time being)
+    if [ "$MB_PYTHON_VERSION" == "2.7" ]; then
+        TOXENV="$TOXENV-tx"
+    fi
 
     # Install pre-compiled wheel and run tests against it
     tox --installpkg "${wheel}" -e "${TOXENV}"
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -724,9 +724,6 @@ class CFFFontData:
     def close(self):
         self.ttFont.close()
 
-    def getGlyphID(self, name):
-        return self.ttFont.getGlyphID(name)
-
     def isCID(self):
         return hasattr(self.topDict, "FDSelect")
 
@@ -859,7 +856,8 @@ class CFFFontData:
 
         return fdDict
 
-    def getfdIndex(self, gid):
+    def getfdIndex(self, name):
+        gid = self.ttFont.getGlyphID(name)
         return self.topDict.FDSelect[gid]
 
     def getfdInfo(self, allow_no_blues, noFlex, vCounterGlyphs, hCounterGlyphs,
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -671,7 +671,7 @@ class CFFFontData:
             font = TTFont(path)
             if "CFF " not in font:
                 raise FontParseError("OTF font has no CFF table <%s>." % path)
-        elif font_format != "PFC":
+        else:
             # Else, package it in an OTF font.
             if font_format == "CFF":
                 with open(path, "rb") as fp:
@@ -689,9 +689,6 @@ class CFFFontData:
             font = TTFont()
             font['CFF '] = newTable('CFF ')
             font['CFF '].decompile(data, font)
-        else:
-            raise NotImplementedError("%s font format is not supported." %
-                                      self.font_format)
 
         self.ttFont = font
         self.cffTable = font["CFF "]
@@ -917,10 +914,14 @@ class CFFFontData:
             fdGlyphDict, fontDictList, finalFDict = fdTools.parseFontInfoFile(
                 fontDictList, fontInfoData, glyphList, maxY, minY,
                 self.getPSName())
+            if hasattr(topDict, "FDArray"):
+                private = topDict.FDArray[fdIndex].Private
+            else:
+                private = topDict.Private
             if finalFDict is None:
                 # If a font dict was not explicitly specified for the
                 # output font, use the first user-specified font dict.
-                fdTools.mergeFDDicts(fontDictList[1:], topDict.Private)
+                fdTools.mergeFDDicts(fontDictList[1:], private)
             else:
-                fdTools.mergeFDDicts([finalFDict], topDict.Private)
+                fdTools.mergeFDDicts([finalFDict], private)
         return fdGlyphDict, fontDictList
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -12,7 +12,12 @@ import re
 
 from fontTools.misc.psCharStrings import T2OutlineExtractor, SimpleT2Decompiler
 from fontTools.misc.py23 import bytechr, byteord, open
-from fontTools.ttLib import TTFont, getTableClass
+from fontTools.t1Lib import T1Font
+from fontTools.ttLib import TTFont, newTable
+from fontTools.cffLib import (CharStrings, GlobalSubrsIndex, PrivateDict,
+                              TopDictIndex, TopDict)
+from fontTools.pens.t2CharStringPen import T2CharStringPen
+from fontTools.pens.recordingPen import RecordingPen
 
 from . import fdTools, FontParseError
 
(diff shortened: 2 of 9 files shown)


tests/integration/test_cli.py
```diff
@@ -93,14 +93,39 @@ def test_multi_reference_is_duplicated(tmpdir):
         autohint(inpaths + ['-o'] + outpaths + ['-r', reference])
 
 
+tx_found = False
+try:
+    subprocess.call(["tx", "-v"])
+    tx_found = True
+except OSError:
```
miguelsousa: I think `subprocess.CalledProcessError` should be included here as well, in case `tx` is found but does not return `0`. So something like this,
```python
except (subprocess.CalledProcessError, OSError):
```

tests/integration/test_hint.py
```diff
@@ -71,14 +72,45 @@ def test_cff(cff, tmpdir):
                    str(tmpdir / basename(out)) + ".xml"])
 
 
-@parametrize("path", glob.glob("%s/dummy/font.p*" % DATA_DIR))
-def test_type1(path, tmpdir):
+tx_found = False
+try:
+    subprocess.call(["tx", "-v"])
```
miguelsousa: Same changes as `test_cli.py`

tests/integration/test_hint.py
```diff
@@ -71,14 +72,45 @@ def test_cff(cff, tmpdir):
...
+def test_type1_cid_raises(path, tmpdir):
     out = str(tmpdir / basename(path)) + ".out"
     options = Options(path, out)
     with pytest.raises(NotImplementedError):
         hintFiles(options)
 
 
+@parametrize("path", glob.glob("%s/dummy/font.pf[ab]" % DATA_DIR))
+@pytest.mark.skipif(tx_found is False, reason="'tx' is missing")
+def test_type1_supported(path, tmpdir):
```
miguelsousa: What is this test actually doing? I'm confused by the two calls to `tx` below; I'd expect to be one call to tx, which would be diff'ed against expected results.

khaledhosny:
Codacy hates subprocess apparently…

miguelsousa:
> Codacy hates subprocess apparently…

I've removed those patterns. Codacy should stop complaining about using `subprocess`.

tests/integration/test_cli.py
khaledhosny: I was reluctant to do this, since many AFDKO tools (used to?) return non-0 for help options. But lets do it and see.

tests/integration/test_hint.py
khaledhosny: It was supposed to dump input and output files and compares them, but now I see that there was a typo in the second call and both ended up dumping the input font, and I was wondering how the tests passed from the first time, LOL.

miguelsousa: I don't know. What exactly are the input and output files that the test is running on?

tests/integration/test_cli.py
```diff
@@ -93,14 +93,39 @@ def test_multi_reference_is_duplicated(tmpdir):
         autohint(inpaths + ['-o'] + outpaths + ['-r', reference])
 
 
+tx_found = False
+try:
+    subprocess.call(["tx", "-v"])
```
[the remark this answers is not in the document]
khaledhosny: Done.

tests/integration/test_hint.py
khaledhosny: The input is the Type 1 fonts under `tests/integration/data/dummy/`; `font.pfa`, `font.pfb` and `font.ps`. The output is what psautohint generates, which here will be roundtripped t1 → cff → t1 by tx, and I think that is the source of the difference.

miguelsousa: OK. The differences are gone now, after you round-tripped the input fonts with tx, yes?

python/psautohint/otfFont.py
```diff
@@ -652,24 +654,37 @@ def convertBezToT2(bezString):
...
-            # It is s CFF font, package it in an OTF font.
-            with open(path, "rb") as fp:
-                data = fp.read()
+            # Else, package it in an OTF font.
+            if font_format == "CFF":
+                with open(path, "rb") as fp:
+                    data = fp.read()
+            else:
+                with tempfile.NamedTemporaryFile() as fp:
+                    _run_tx(["-cff", "+b", "-std", path, fp.name], font_format)
```
miguelsousa: Enable `USE_TX` tests on AppVeyor as well. I want to make sure this call does not fail on Windows.
It won't hurt to enable them on Linux as well, since we've experienced quite interesting results on the afdko repository when using temporary files.

tests/integration/test_hint.py
miguelsousa: That test font is not that good, since it only contains one `FDArray` (it has been like that since it was added to the repo). Let me craft a better test font for that format.

miguelsousa: Added new **font.ps**: https://github.com/adobe-type-tools/psautohint-testdata/commit/5febaa3f9c98cbd4bf5c1da0a078883600cb7987

python/psautohint/otfFont.py
```diff
@@ -652,24 +654,37 @@ def convertBezToT2(bezString):
     return t2Program
 
 
+def _run_tx(args, font_format):
+    try:
+        subprocess.check_call(["tx"] + args)
+    except subprocess.CalledProcessError as e:
+        raise FontParseError(e)
+    except OSError:
```
miguelsousa: Thinking about this some more, I think this `except` should be moved to a new method (`_check_tx`?) that is called much earlier, preferably in `__main__.get_options`, right after `_validate_font_paths` is called. Then we can simply throw a `parser.error`.
Also, I think the error message should say something about how to get **tx**.

`NotImplementedError` (in **__main__.py**) won't be necessary anymore once the changes above are made.

python/psautohint/otfFont.py
khaledhosny: Enabled on AppVeyor. I tried installing AFDKO on Linux initially but it failed due to some missing headers, did not investigate.

anthrotype: Yeah, I also noticed NamedTemporaryFile doesn't always work with Windows. Sometimes I have to create one with delete=False, then close it, and then finally remove it manually after doing my things with it. Otherwise Windows seems to lock the open temp file and doesn't let the other subprocess write to it. Or I just skip NamedTemporaryFile use the low level tempfile.mkstemp directly.

anthrotype: > I tried installing AFDKO on Linux initially but it failed due to some missing headers

because there are no manylinux1 wheels on PyPI so one has to build from source and you need the gcc-multilib to be able to compile 32 bit on a 64 bit host.

khaledhosny: I thought there is one now after the recent changes (I’m using `--pre`), but I guess I’m mistaken. 32 bit build is needed for old `autohintexe`, right? We could wait then until this one is really dead a buried.

anthrotype: ah you're right there is a manylinux1 wheel, at least for 2.7 and 3.6:
https://pypi.org/project/afdko/2.8.0rc0/#files

python/psautohint/otfFont.py
```diff
@@ -652,24 +654,38 @@ def convertBezToT2(bezString):
...
+            if font_format == "CFF":
+                with open(path, "rb") as fp:
+                    data = fp.read()
+            else:
+                fd, temp_path = tempfile.mkstemp()
+                os.close(fd)
+                _run_tx(["-cff", "+b", "-std", path, temp_path])
+                with open(temp_path, "rb") as fp:
+                    data = fp.read()
+                os.remove(temp_path)
```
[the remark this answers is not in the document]
miguelsousa: I don't think we need to bother removing a temporary file that is properly written in the system's temporary folder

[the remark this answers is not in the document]
anthrotype: a `try: ... finally:` clause is just good practice and it's what NamedTemporaryFile would do automatically upon closing the file (if Windows didn't prevent us from using it).

khaledhosny:
I don’t even need `os.pwrite()`, I read it `os.bwrite()` and though that the `b` is for bytes :smile:

python/psautohint/otfFont.py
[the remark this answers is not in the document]
khaledhosny: @miguelsousa I’d rather prefer to properly delete the temporary files, I have seen software that do not do so in the past and one easily ends with thousands of files in the temp directory, and this can sometimes be an issue.

python/psautohint/otfFont.py
khaledhosny: Hmm, installing afdko seems to be silently skipped https://travis-ci.org/adobe-type-tools/psautohint/jobs/424703851#L1105, @anthrotype any idea?

python/psautohint/otfFont.py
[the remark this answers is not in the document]
miguelsousa: 👍

config.sh
```diff
@@ -6,7 +6,7 @@
 function pre_build {
     # Any stuff that you need to do before you start building the wheels
     # Runs in the root directory of this repository.
-    if [ "${USE_TX}" = true ]; then
+    if [ \( -n "$IS_OSX" -a "$MB_PYTHON_VERSION" = "2.7" \) -o \( "$MB_PYTHON_VERSION" = "3.6" \) ]; then
```
anthrotype: ugly, but it should work. You didn't like the `env_vars.sh` thinghy?

khaledhosny: Not that I didn’t like it, I just still don’t understand how to get it to work. Anyway, the or branch didn’t work, so I’m just installing it for any 3.6 Python, on AppVeyor we check 2.7 anyway.

config.sh
```diff
@@ -6,7 +6,7 @@
 function pre_build {
     # Any stuff that you need to do before you start building the wheels
     # Runs in the root directory of this repository.
-    if [ \( -n "$IS_OSX" -a "$MB_PYTHON_VERSION" = "2.7" \) -o \( "$MB_PYTHON_VERSION" = "3.6" \) ]; then
+    if [ "$MB_PYTHON_VERSION" = "3.6" ]; then
```
anthrotype: it appears that `MB_PYTHON_VERSION` is passed on to the docker container as `PYTHON_VERSION` (without the `MB_` prefix). However, on the macOS multibuild environment, we don't have a `PYTHON_VERSION` variable, only an `MB_PYTHON_VERSION`, so to cover both you need to ... check both?
https://github.com/matthew-brett/multibuild/blob/548ebc80443250ead493164cba330e56b3558c70/travis_linux_steps.sh#L86

or just use `env_vars.sh`

sorry... :(

config.sh
khaledhosny: This is absurd, even `MB_PYTHON_VERSION` is not getting passed to this function on Linux. I’m lost.

config.sh
anthrotype: you can see that it didn't catch the `MB_PYTHON_VERSION` (so tx isn't installed), otherwise something about pip installing tx would appear before this line:
https://travis-ci.org/adobe-type-tools/psautohint/jobs/424737479#L526

config.sh
anthrotype: It is annoying. I'll send a patch to upstream multibuild to fix this.

khaledhosny: OK, afdko installed but still tx is not found, grrr.

config.sh
anthrotype: oh, I'm so sorry Khaled I've misled you... So, the problem is that, it *is* now installing the afdko in the build environment (inside the CentOS docker) where the wheel is getting built. However, you want it inside the *test* environment, where you run the tests!
The latter is also a docker image (a vanilla Ubuntu Trusty). To configure the packages that gets installed in the test environment you have to set the `TEST_DEPENDS` environment variable.

We already use that in our .travis.yml, to install from dev-requirements.txt. To fix this you could add the `afdko` to dev-requirements.txt.

You could use an environment marker to do it conditionally only when `afdko; python_version == '3.6'` but writing that in dev-requirements.txt may be misleading..

Or we could make a tox environment where we install afdko and use that TOXENV...

anthrotype:
ok I think we nailed the tox with conditional tx stuff. Let me know if you want me to squash my commits.

khaledhosny:
Please squash the commits, and enable py27-cov-tx on Linux as well. Thanks again!

anthrotype:
squashed last two commits

miguelsousa:
👏

Issue #101, opened 2018-09-05
New stable release


khaledhosny:
I think it is time to cut a new stable release, @miguelsousa and @anthrotype any objections?

Given the nature of changes, I think it should be 1.8.0 not 1.7.1 as I originally planned.

anthrotype:
sounds good to me.

miguelsousa:
Sounds good to me too

anthrotype:
Ah! Something just occurred to me.. since we now install a autohintexe console script, what if one installs afdko after psautohint? The latter’s “good” autohintexe launcher would be overwritten by the “bad” 32-bit only standalone executable. I think that if we want the two to coexist peacefully in the same site packages we need afdko to stop installing a autohintexe to bin/Scripts.

anthrotype:
Similarly uninstalling afdko would remove the autohintexe launcher installed by psautohint..

khaledhosny:
Right, so I guess we need AFDKO to drop its autohintexe first, cut a new stable release (after whatever number of beta releases until we are confident enough, of course), then we immediately cut one (so that people installing AFDKO would still get a working autohint), right?

khaledhosny:
Alternatively, we can drop the wrapper (not a fan of it anyway), and AFDKO’s autohint could be adapted to look for the binary in the right place not in the PATH. That would be the cleaner solution I think.

anthrotype:
I like the second option

khaledhosny:
I just made a new stable release.

PR #102, opened 2018-09-05
Drop autohintexe wrapper

khaledhosny:
See https://github.com/adobe-type-tools/psautohint/issues/101#issuecomment-418846200

### DIFF
--- a/python/psautohint/__main__.py
+++ b/python/psautohint/__main__.py
@@ -15,7 +15,7 @@ import sys
 
 from fontTools.misc.py23 import open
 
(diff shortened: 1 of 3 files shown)


khaledhosny:
In `__init__.py`.

miguelsousa:
Is it needed? I'd rip that out as well

khaledhosny:
I’m keeping it to be used by AFDKO, see https://github.com/adobe-type-tools/afdko/pull/596, not sure if we can move the definition to AFDKO (not in its current form that relies on `__file__` at least).

anthrotype:
I think it should stay here

PR #113, opened 2018-09-28
Try to fix signed integer overflow

khaledhosny:
Signed integer overflow is undefined behaviour, UBSAN gives:

../libpsautohint/src/flat.c:191:21: runtime error: signed integer overflow: -51456 * 43008 cannot be represented in type 'int'

### DIFF
--- a/libpsautohint/src/flat.c
+++ b/libpsautohint/src/flat.c
@@ -179,6 +179,7 @@ FMiniFltn(Cd f0, Cd f1, Cd f2, Cd f3, FltnRec* pfr)
         if (inbbox) {
             Fixed eqa, eqb, x, y;
             Fixed EPS;
+            /* int64_t instead of Fixed to avoid integer overflow below */
             int64_t d;
             x = c0x;
             y = c0y;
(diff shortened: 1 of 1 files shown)


khaledhosny:
I changed the variable holding the result to be `int64_t` as well.

libpsautohint/src/flat.c
```diff
@@ -178,7 +178,8 @@ FMiniFltn(Cd f0, Cd f1, Cd f2, Cd f3, FltnRec* pfr)
         }
         if (inbbox) {
             Fixed eqa, eqb, x, y;
-            Fixed EPS, d;
+            Fixed EPS;
+            int64_t d;
```
cjchapman: Might be good to add a comment like "needs to be int64_t instead of Fixed to prevent overflow" here.

cjchapman: (Sorry for the late suggestion, OK with me to leave as is.)

khaledhosny: Done.

cjchapman: 👍

Issue #105, opened 2018-09-08
Support CFF2 for VF hinting


madig:
```
$ psautohint '/home/nikolaus/Downloads/SourceSansVariable-Roman(1).otf'
ERROR: Unhandled exception occurred
Traceback (most recent call last):
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/__main__.py", line 707, in main
    hintFiles(options)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 212, in hintFiles
    hintFile(options, path, outpath, reference_master=False)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 221, in hintFile
    fontData = openFile(path, options)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 200, in openFile
    font = CFFFontData(path, font_format)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/otfFont.py", line 673, in __init__
    raise FontParseError("OTF font has no CFF table <%s>." % path)
psautohint.FontParseError: OTF font has no CFF table </home/nikolaus/Downloads/SourceSansVariable-Roman(1).otf>.
Traceback (most recent call last):
  File "/tmp/venv/bin/psautohint", line 11, in <module>
    sys.exit(main())
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/__main__.py", line 707, in main
    hintFiles(options)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 212, in hintFiles
    hintFile(options, path, outpath, reference_master=False)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 221, in hintFile
    fontData = openFile(path, options)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/autohint.py", line 200, in openFile
    font = CFFFontData(path, font_format)
  File "/tmp/venv/lib64/python3.6/site-packages/psautohint/otfFont.py", line 673, in __init__
    raise FontParseError("OTF font has no CFF table <%s>." % path)
psautohint.FontParseError: OTF font has no CFF table </home/nikolaus/Downloads/SourceSansVariable-Roman(1).otf>.
```

khaledhosny:
[...]

PR #116, opened 2018-10-13
Add psstemhist tool

khaledhosny:
This ports relevant code from AFDKO’s `stemhist` and adds a similar tool. This also allows, again, to use `autohintexe` binary instead of the Python C extension.

The `psstemhist` tool uses `autohintexe` by default since we get mysterious crashes when running the test suite while using C extension for stemhist. I tried to debug these crashes for weeks but no avail, I’m suspecting that something is corrupting the memory and causing subsequent calls to the library to crash, but that is just speculation.

### DIFF
--- a/python/psautohint/__init__.py
+++ b/python/psautohint/__init__.py
@@ -3,11 +3,14 @@ from __future__ import print_function, absolute_import
 import os
 import sys
 import plistlib
+import logging
 
 from fontTools.misc.py23 import tobytes, tounicode
 
 from . import _psautohint
 
+
+log = logging.getLogger(__name__)
 
 __version__ = _psautohint.version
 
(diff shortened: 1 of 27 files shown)


khaledhosny:
Travis is failing on macOS due to some unclear permission issue, but @anthrotype couldn’t reproduce it locally, not sure what to do now.

python/psautohint/__init__.py
```diff
@@ -3,15 +3,28 @@
...
+
 __version__ = _psautohint.version
 
 
+AUTOHINTEXE = os.path.join(
+    os.path.dirname(__file__),
+    "autohintexe" + (".exe" if sys.platform in ("win32", "cygwin") else "")
+)
+if not os.path.isfile(AUTOHINTEXE) or not os.access(AUTOHINTEXE, os.X_OK):
+    log.warning(
```
anthrotype: it's better if you use `warnings.warn` here, instead of logging, as the latter might not be configured by the client application, and this warning is really directed not to the end-user of the tool, but to the developer integrating the psautohint library into an application, or to the system admin deploying and installing the package.
`warnings` is more about issues with the code itself, whereas logging more about what the code does

python/psautohint/__main__.py
```diff
@@ -717,5 +734,180 @@ def main(args=None):
...
+def stemhist(args=None):
+    options, pargs = get_stemhist_options(args)
+
+    try:
+        hintFiles(options)
+    except Exception as ex:
+        if pargs.traceback:
+            raise
+        logging.error(ex)
+        sys.exit(1)
```
anthrotype: it's better that you don't raise SystemExit inside here or similar main-like functions that are used as console scripts entry points. Just return the exit code `return 1`, and the wrapper will raise `sys.exit(your_entry_point())` for you. Otherwise when you call stemhist in the test suite, and it fails, it will stop the whole test run and pytest process will exit abruptly (see `test_cli.py::test_stemhist` where you call `stemhist` function without catching SystemExit)

anthrotype:
It appears to be an issue with [delocate-wheell](https://github.com/matthew-brett/delocate) command, that is run by multibuild after generating the wheel file.
I've opened an issue upstream
https://github.com/matthew-brett/delocate/issues/42

anthrotype:
I found the issue and sent the patch upstream 
https://github.com/matthew-brett/delocate/pull/43

temporarily using my fork of delocate seems to fix the issue (d5fd091)

python/psautohint/__main__.py
khaledhosny: Aha, that is why the logs were truncated in the failing builds. Thanks!

khaledhosny:
Thanks @anthrotype for debugging and fixing the permission issue.

python/psautohint/__init__.py
khaledhosny: Done, thanks!

khaledhosny:
> There's quite a bit of argparse-related duplicate code in `__main__.py` now. It'd be nice to consolidate it in a future round.

I didn’t like that either and my original plane was to add an option to `psautohint` that makes it do the stem analysis, but it I felt it is going to complicate option parsing.

> I don't know how critical this is nor if it's even related to the changes in this PR, but I noticed this warning when I compiled from source:
> ```
> libpsautohint/src/charpath.c:1776:5: warning: code will never be executed [-Wunreachable-code]
>    WriteToBuffer("%% %s\n", gGlyphName);
> ```

It is been like that since the MM code was resuscitated. That bit of code is for merging glyphs to produce variable ones, but we don’t currently use it and I kept the code not commented out so it does not bitrot, we might need it for #105.

Issue #103, opened 2018-09-07
segfault in control.c::GetNewPtLst when compiled with -O3


anthrotype:
I compile my python from source using pyenv. By default, this python gets compiled with -O3 flags. The python interpreter's CFLAGS are passed on by distutils when compiling any extension modules.
Now, when the psautohint module is compiled with -O3, I get a segmentation fault at this line
https://github.com/adobe-type-tools/psautohint/blob/master/libpsautohint/src/control.c#L351

This is triggered for example when running integration/test_hints.py

If I force compile psautohint module with CFLAGS=-O2 the segfaults goes away.

I have no idea why.

mdecerbo:
FWIW, I can confirm the same issue is happening with commit cde9b1856. If I compile with `-O3` I get the segfault below:
```
#0  GetNewPtLst () at libpsautohint/src/control.c:351
#1  XtraHints (e=0x7fffeb77ff20 <vm+16320>) at libpsautohint/src/control.c:367
#2  0x00007fffeb571746 in StartNewHinting (e=e@entry=0x7fffeb77ff20 <vm+16320>, 
    hLst=hLst@entry=0x7fffeb784e64 <vm+36612>, vLst=vLst@entry=0x0)
    at libpsautohint/src/auto.c:671
#3  0x00007fffeb5731ea in AutoExtraHints (movetoNewHints=false)
    at libpsautohint/src/auto.c:1093
#4  0x00007fffeb56086d in AddHintsInnerLoop (extrahint=<optimized out>, 
    srcglyph=<optimized out>) at libpsautohint/src/control.c:802
#5  AddHints (extrahint=<optimized out>, 
    srcglyph=0x11463e0 "% uni2619\nsc\n830 123 mt\n830 110 828 98 823 88 ct\n795 41 759 30 747 27 ct\n739 46 732 67 732 86 ct\n732 127 775 148 775 181 ct\n775 210 759 212 749 212 ct\n740 212 730 210 722 207 ct\n735 230 742 232 774 23"...)
    at libpsautohint/src/control.c:879
```
This is when building commit f87af5d1 of [Libertinus](https://github.com/alif-type/libertinus) with an up-to-date FontForge. Python 3.6.8, gcc 4.8.5. Using `CFLAGS="-O2"` eliminates the problem.

josh-hadley:
@mdecerbo as an aid in diagnosing this, could you paste the result of the following (using the Python where you're experiencing the segfault):
```
[...]

PR #121, opened 2018-10-25
Fix some compiler warnings

khaledhosny:


anthrotype:
This seems to fix #103
I checked out this branch, run `tox -e py37-cov` using my pyenv-compiled python 3.7 that uses -O3, and the all tests passed, no more segfaults. Thanks @khaledhosny!

PR #128, opened 2018-11-06
Port to fontTools.ufolib

khaledhosny:


python/psautohint/__init__.py
```diff
@@ -31,18 +30,19 @@ class FontParseError(Exception):
...
-            keys = ('creator', 'formatVersion')
-            if metainfo and all([key in metainfo for key in keys]):
-                return True
-    return False
+    from fontTools.ufoLib import UFOReader
+    from fontTools.ufoLib.errors import UFOLibError
+    try:
+        UFOReader(path, validate=False)
+        return True
+    except (UFOLibError, KeyError, TypeError):
```
khaledhosny: @anthrotype Having to catch `KeyError` and `TypeError` here is a bit ugly I think, not sure if I should have catch `Exception` instead. `UFOReader.readMetaInfo()` tries to access items in the parsed dict unconditionally, but it can be None (empty `metainfo.plist`) or misses `formatVersion`, I’d have expected `UFOLibError` in these cases.

python/psautohint/ufoFont.py
```diff
@@ -325,8 +326,8 @@
 
 ADOBE_DOMAIN_PREFIX = "com.adobe.type"
 
-PROCESSED_LAYER_NAME = "AFDKO ProcessedGlyphs"
-PROCESSED_GLYPHS_DIRNAME = "glyphs.%s.processedGlyphs" % ADOBE_DOMAIN_PREFIX
+PROCESSED_LAYER_NAME = "%s.processedglyphs" % ADOBE_DOMAIN_PREFIX
```
anthrotype: The layer name can be anything, it’s the directory name that is recommended to be lowercase and unique etc.

python/psautohint/__init__.py
khaledhosny: No it doesn’t: https://github.com/fonttools/fonttools/blob/76c85e3c3b6f2fa0fa4a8bdec7036b4bb06c6154/Lib/fontTools/ufoLib/__init__.py#L392-L395

anthrotype: Ok, we need to fix that.
An except Exception: is not that bad in this particular case, but this also works

python/psautohint/ufoFont.py
khaledhosny: I just followed https://github.com/adobe-type-tools/afdko/commit/a4aa4762bdcd4d4012226c766f6ca27ddb0b68a4, these need to be the same between afdko and psautohint.

miguelsousa: @anthrotype it can be anything BUT the ufonormalizer derives the **folder name** from the **layer name**, so the least disruptive solution is to adjust the name of the layer.

anthrotype: i think it does that only when the directory name contains a non-lowercase character, but anyway as you prefer.

Issue #122, opened 2018-10-28
Some outlines are modified when using --reference-font


madig:
Using `-r` to hint compatibly results in some modified outlines in the non-reference masters. This prevents building a variable font.

1. `curl -O https://raw.githubusercontent.com/madig/cantarell-fonts/master/src/Cantarell.glyphs`
2. `fontmake -g Cantarell.glyphs -o otf --keep-overlaps`
3. Look at e.g. the brevecomb in uni1EAE in all masters, they should be compatible.
4. `psautohint -r master_otf/Cantarell-Regular.otf master_otf/Cantarell-Bold.otf master_otf/Cantarell-Light.otf`
5. Check again. The path is as it should be in Regular but modified in Light and Bold.

khaledhosny:
Looks like a memory issue, here I get:
```
double free or corruption (!prev)
Aborted (core dumped)
```

Investigating.

PR #130, opened 2018-11-15
Off by one error when writing MM out buffer

khaledhosny:
Found while investigating #122.


madig:
Does this just plug the memory hole or also prevent outline modification?

khaledhosny:
I missed the outline modification part, I thought you were getting *corrupt* outlines and assume it was caused by the memory issue, but now I see what you are talking about, and no it doesn’t fix it. I better get some sleep now :smile:

PR #131, opened 2018-11-29
Fixes for #122

khaledhosny:
1. Rewrite autohint.hintFiles() so that the output of hint_compatible_bez_glyphs() is used for reference master as well.
2. Stop adding short line segments to “zero length closepaths”.

### DIFF
--- a/libpsautohint/autohintexe.c
+++ b/libpsautohint/autohintexe.c
@@ -79,39 +79,32 @@ printHelp(void)
     fprintf(stdout, "   -v print versions.\n");
 }
 
-#define STR_LEN 200
-#define WRITE_TO_BUFFER(text, a1, a2, a3)                                      \
-    if (userData) {                                                            \
-        char str[STR_LEN];                                                     \
-        ACBuffer* buffer = (ACBuffer*)userData;                                \
-        int len = snprintf(str, STR_LEN, text, a1, (double)a2, (double)a3);    \
-        if (len > 0)                                                           \
-            ACBufferWrite(buffer, str,                                         \
-                          (size_t)(len <= STR_LEN ? len : STR_LEN));           \
-    }
-
 static void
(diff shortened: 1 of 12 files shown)


miguelsousa:
@khaledhosny the checks didn't pass

khaledhosny:
Yes, some tests are crashing on the CI machines but not on mine, scratching my head trying to figure out how to debug this…

khaledhosny:
OK, I know knows what is crashing, but no idea why. Still investigation, do not merge.

khaledhosny:
Still not able to reproduce the crashes locally. I even used [Travis Docker image](https://docs.travis-ci.com/user/common-build-problems/#troubleshooting-locally-in-a-docker-image) hoping to replicate the same environment, but no luck.

khaledhosny:
@anthrotype Can you help here? You were able to reproduce similar weird crashes in the past, any chance you can try this branch locally and see if you can reproduce the crashes (they are transient, but no matter how many times I run this locally it doesn’t crash for me).

khaledhosny:
@miguelsousa This should be ready to review once the build finishes. I couldn’t fix the crash (still can’t reproduce it), but I worked around it by partially reverting to the old approach.

anthrotype:
@khaledhosny hey, I was away last week. Do you still need me to try reproduce this locally? Should I revert any commits to repro, if so which ones?

khaledhosny:
Yes, the crash is not fixed yet. Try reverting the last commit on master and see if you can reproduce it.

anthrotype:
I reverted the e09f397ce77a4a7ef9b4afeb987dc5dc9925c42b commit, and run `tox -e py37-cov` using a pyenv-compiled Python 3.7.1 running on my Debian Linux box. All tests passed without crashes.

khaledhosny:
The crashes are flaky (the tests didn’t always crash on CI), try running it a few more times if possible,

PR #133, opened 2018-12-06
Tweaks to fix hashmap differences between py2 & py3

miguelsousa:
While working on switching **makeinstancesufo** to using psautohint (instead of autohint) ran into py2/py3 differences in the `com.adobe.type.processedHashMap` file.
These changes can be reverted when we stop supporting py2.

### DIFF
--- a/python/psautohint/ufoFont.py
+++ b/python/psautohint/ufoFont.py
@@ -121,7 +121,7 @@ import shutil
 
 from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


miguelsousa:
@khaledhosny I'm going to issue a dot release, OK? Need these changes to wrap up the work on makeinstancesufo.

khaledhosny:
Sure, no problem.

PR #143, opened 2019-02-15
Don't hard code the name of the default layer

### LINKED ISSUE #135: [psautohint] modifies layercontents.plist
`psautohint` renames the foreground layer in a UFO3’s `layercontents.plist` file from `foreground` to `public.default`. In addition to that, all layers that are not the foreground or `com.adobe.type.processedglyphs` are removed from the file.

The good news is that the glyphs are not lost, only their wiring (which still isn’t optimal).

```
diff --git a/test.ufo/layercontents.plist b/test.ufo/layercontents.plist
index 84fa544..b076808 100644
--- a/test.ufo/layercontents.plist
+++ b/test.ufo/layercontents.plist
@@ -3,20 +3,12 @@
 <plist version="1.0">
   <array>
     <array>
-      <string>foreground</string>
+      <string>public.default</string>
       <string>glyphs</string>
     </array>
     <array>
-      <string>background</string>
-      <string>glyphs.background</string>
-    </array>
-    <array>
-      <string>other layer</string>
-      <string>glyphs.other layer</string>
-    </array>
-    <array>
-      <string>condensed</string>
-      <string>glyphs.condensed</string>
+      <string>com.adobe.type.processedglyphs</string>
+      <string>glyphs.com.adobe.type.processedglyphs</string>
     </array>
   </array>
 </plist>
```

Test UFO file is attached.
[test.ufo.zip](https://github.com/adobe-type-tools/afdko/files/2687175/test.ufo.zip)

miguelsousa:
Fixes #135

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ dist
 .coverage*
 htmlcov/
 libpsautohint/src/version.h
+pip-wheel-metadata/psautohint.dist-info
--- a/python/psautohint/ufoFont.py
+++ b/python/psautohint/ufoFont.py
@@ -124,8 +124,7 @@ from collections import OrderedDict
 from fontTools.misc.py23 import SimpleNamespace, open, tostr, round
 from fontTools.pens.basePen import BasePen
 from fontTools.pens.pointPen import AbstractPointPen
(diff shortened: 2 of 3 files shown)


miguelsousa:
@khaledhosny not sure where to add the test case for this. Can you take care of it? Thanks

khaledhosny:
I’ll take a look.

PR #145, opened 2019-02-19
Add support for sparse sources

readroberts:
Add support for a set of variable font sources, where the source fonts have different subsets of the default font glyph set.


miguelsousa:
LGTM

PR #150, opened 2019-03-14
Fix problem with MM hint lists sorting differently.

readroberts:
Even though the MM bez files are merge compatible after hinting, the T2 CharStrings built from them may not be compatible. This is because hint pairs in T2 CharStrings must be sorted, and duplicates must be eliminated. The differing values of hint pairs between the MM source fonts may cause them to be sorted and filtered differently. The fix was to define a list of the selected hint pair indices and order from the reference font, and then apply this list in the region fonts.


readroberts:
Established new branch

PR #151, opened 2019-03-14
Fix problem with MM hint lists sorting differently.

readroberts:
Even though the MM bez files are merge compatible after hinting, the T2 CharStrings built from them may not be compatible. This is because hint pairs in T2 CharStrings must be sorted, and duplicates must be eliminated. The differing values of hint pairs between the MM source fonts may cause them to be sorted and filtered differently. The fix was to define a list of the selected hint pair indices and order with the reference font, and then apply this list in the region fonts.

### DIFF
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -15,6 +15,7 @@ import tempfile
 from fontTools.misc.psCharStrings import T2OutlineExtractor, SimpleT2Decompiler
 from fontTools.misc.py23 import bytechr, byteord, open
 from fontTools.ttLib import TTFont, newTable
+from collections import namedtuple
 
 from . import fdTools, FontParseError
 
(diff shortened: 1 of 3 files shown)


python/psautohint/otfFont.py
```diff
@@ -634,15 +710,35 @@ def convertBezToT2(bezString):
...
+                pos = hintMask.listPos
+                hBytes = hintMask.maskByte(hhints, vhints)
+                t2List[pos] = [["hintmask"], hBytes]
+                if in_mm_hints:
+                    hm = RefHintMask(pos=pos, hbytes=hBytes)
+                    mm_hint_info.hint_masks.append(hm)
+        elif (mm_hint_info is not None):
+            # This is a MM region font: apply hint masks from reference font.
+            hm0_mask = mm_hint_info.hint_masks[0].hbytes
+            if type(t2List[0][0]) == HintMask:
```
miguelsousa: Use `isinstance()` instead of `type()`, as suggested by Codacy

readroberts: Done.

miguelsousa:
@readroberts is it feasible to add a test that covers these changes?

readroberts:
@msousa. There already is. The test function test_mhint.py::test_sparse_mmotf() covers this.

python/psautohint/otfFont.py
```diff
@@ -440,17 +437,70 @@ def makeRelativeCTArgs(argList, curX, curY):
...
+    # should filter out duplicates
+
+    numHints = len(hints)
+    index_list = list(range(numHints))
+    hint_list = list(zip(hints, index_list))
+    hint_list.sort()
+    new_hints = [hint_list[i] for i in range(1, numHints)
+                 if hint_list[i][0] != hint_list[i-1][0]]
+    new_hints = [hint_list[0]] + new_hints
+    hints, hint_order = list(zip(*new_hints))
```
blueshade7: I believe the comment means: "calling hint_order.index(index_from_bez) will yield the hint index after sorting".
I first interpreted it backwards: "hint_order[index_from_bez] will yield the hint index after sorting".
Seems misleading; better clarify?

python/psautohint/otfFont.py
```diff
@@ -499,33 +549,45 @@ def convertBezToT2(bezString):
             lastPathOp = token
         elif token == "rb":
             lastPathOp = token
-            try:
-                i = hhints.index(argList)
-            except ValueError:
-                i = len(hhints)
+            if in_mm_hints:
```
blueshade7: The same logic is repeated four times. Maybe better to factor it out as a function?

python/psautohint/otfFont.py
readroberts: Will do this, and your other suggestion. FYI, I am working on a new commit to remove hint pairs that are incorrect in the MM hinted region fonts.

PR #155, opened 2019-04-18
Fix mm mask issues

readroberts:
Added to code to 1) apply hint masks and counter masks from default font to other fonts 2) remove hints that are not valid in all fonts, and 3) correct hint and counter masks when invalid hints are removed. 
Fixed bugs in C library for MM hinting


python/psautohint/autohint.py
```diff
@@ -657,18 +683,34 @@ def hint_compatible_fonts(options, fonts, paths, glyph_list, glyphs,
...
+        bez_glyphs = [g[name].bez_data for g in glyphs]
         new_bez_glyphs = hint_compatible_glyphs(options, name, bez_glyphs,
                                                 masters, fontinfo)
-
         if options.logOnly:
             continue
 
-        hinted[name] = (new_bez_glyphs, widths)
+        if reference_font is None:
+            fonts = [g[name].font for g in glyphs]
```
blueshade7: These two lines probably equivalent to:
```
reference_font = glyphs[0][name].font
```

python/psautohint/otfFont.py
```diff
@@ -1000,3 +1099,123 @@ def getfdInfo(self, allow_no_blues, noFlex, vCounterGlyphs, hCounterGlyphs,
...
+            hint_program.extend(decompiler.h_hint_args)
+            hint_program.append(op_hstem)
+        if decompiler.v_hint_args:
+            hint_program.extend(decompiler.v_hint_args)
+            # Don't need to append op_vstem, as this is still in hint_program.
+            program = hint_program + program
+
+        # Re-calculate the hint masks.
+        if is_reference_font:
+            h_args = decompiler.h_hint_args
```
blueshade7: The following piece of code for hhints is repeated for vhints and they look the same.
Maybe we may want to factor it out as a function?

python/psautohint/otfFont.py
```diff
@@ -456,35 +455,69 @@ def build_hint_order(hints):
...
+
+def check_hint_pairs(hint_pairs, mm_hint_info, last_idx=0):
+    # pairs must be in ascending order by bottom (or left) edge,
+    # and pairs in a hint group must not overlap.
+
+    # check order first
+    hint_list = [make_abs(hint_pair) for hint_pair in hint_pairs]
+    bad_hint_idxs = set()
+    prev = hint_list[0]
+    for i, hint_pair in enumerate(hint_list[1:], 1):
```
blueshade7: When a pair of stem hints is not in the ascending order, it may not be so trivial to determine which one should be regarded as bad. The code here appears to mark the second one in a pair as bad, but consider four stem hints: A, B, C, D in the correct order, then given a list of: C, A, B, D. I believe this code will regard A and B as bad, but C as bad would minimize the number of bad hints. The same could be said to overlapping hints.

python/psautohint/otfFont.py
```diff
@@ -352,14 +361,14 @@ def makeHintList(hints, needHintMasks, isH):
 kHintArgsMatch = 2
 
 
-def checkStem3ArgsOverlap(argList, hintList):
+def checkStem3ArgsOverlap(arg_list, hint_list):
     # status == 0 -> no overlap
     # status == 1 -> arg are the same
     # status = 2 -> args overlap, and are not the same
     status = kHintArgsNoOverlap
```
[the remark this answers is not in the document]
readroberts: Constant names are correct., comments are outdated and wrong. Removed comments

python/psautohint/otfFont.py
```diff
@@ -376,7 +385,26 @@ def checkStem3ArgsOverlap(argList, hintList):
...
+            # The args match args in this control mask.
+            if overlap_status == kHintArgsMatch:
+                break
+        if overlap_status != kHintArgsMatch:
+            mask = HintMask(0)
+            counter_mask_list.append(mask)
+            dst_hints.extend(arg_list)
+
+
+def build_counter_mask_list(h_stem3_list, v_stem3_list):
```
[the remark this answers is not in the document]
readroberts: Comments are out of date. Fixed.

python/psautohint/otfFont.py
readroberts: Good point. Done.

python/psautohint/otfFont.py
readroberts: @blueshade7 I see your point. Looking at several cases, I think that always removing the first pair will not cause this problem - do you agree?

blueshade7: I cannot be sure without seeing real world examples. Let me assume now it is doing the right thing or at least not a bad thing. Let's revisit when we see a problem later.

PR #164, opened 2019-05-30
Hashing unnormalized floats

miguelsousa:
@khaledhosny the current commits in this PR achieve the opposite of https://github.com/adobe-type-tools/afdko/commit/aff99e04a38e3f8852240c8c46d6179d1f341be9, as we agreed.

However, if you look at the [current hash results](https://github.com/adobe-type-tools/psautohint-testdata/compare/6676a7e0a303e32cf36199be8e386675047a30ab...2dc10194fd88d271538c3b5717acd571adf14bc9#diff-fc17c6b280ff7f55ea51e5787368be8c), there are other values that could also be normalized, like the advance and the outline's points.

I think HashPointPen should normalize all the values. This means that some hashes will not match the ones currently produced by `afdko.ufotools`, but the latter can be updated to stay in sync. I understand that these changes will invalidate some of the hashes in existing fonts, but I'm not concerned about that; my focus in on improving the output of HashPointPen.

What are your thoughts?

### DIFF
--- a/python/psautohint/ufoFont.py
+++ b/python/psautohint/ufoFont.py
@@ -404,6 +404,8 @@ class UFOFontData:
 
         glyph = BezGlyph(bezData)
         glyphset.readGlyph(name, glyph)
+        if hasattr(glyph, 'width'):
+            glyph.width = norm_float(glyph.width)
         self.newGlyphMap[name] = glyph
 
         # updateFromBez is called only if the glyph has been autohinted which
@@ -802,7 +804,7 @@ class HashPointPen(AbstractPointPen):
 
     def __init__(self, glyph):
         self.glyphset = getattr(glyph, "glyphSet", None)
-        self.width = round(getattr(glyph, "width", 1000), 9)
+        self.width = norm_float(round(getattr(glyph, "width", 1000), 9))
         self.data = ["w%s" % self.width]
 
     def getHash(self):
@@ -823,7 +825,8 @@ class HashPointPen(AbstractPointPen):
             pt_type = ""
         else:
             pt_type = segmentType[0]
-        self.data.append("%s%s%s" % (pt_type, repr(pt[0]), repr(pt[1])))
+        self.data.append("%s%s%s" % (
+            pt_type, repr(norm_float(pt[0])), repr(norm_float(pt[1]))))
 
     def addComponent(self, baseGlyphName, transformation, identifier=None,
                      **kwargs):
@@ -831,7 +834,7 @@ class HashPointPen(AbstractPointPen):
 
         for i, v in enumerate(transformation):
             if transformation[i] != self.DEFAULT_TRANSFORM[i]:
-                self.data.append(str(round(v, 9)))
+                self.data.append(str(norm_float(round(v, 9))))
 
         self.data.append("w%s" % self.width)
         glyph = self.glyphset[baseGlyphName]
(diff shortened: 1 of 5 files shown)


miguelsousa:
@khaledhosny WDYT? Normalize everywhere?

khaledhosny:
Yes, makes sense for consistency.

python/psautohint/ufoFont.py
```diff
@@ -831,12 +831,19 @@ def addComponent(self, baseGlyphName, transformation, identifier=None,
...
         glyph = self.glyphset[baseGlyphName]
         glyph.drawPoints(self)
 
+    @staticmethod
+    def norm_float(value):
+        """Converts a float (whose decimal part is zero) to integer"""
+        if isinstance(value, float) and value % 1 == 0:
+            return int(value)
+        return value
+
```
khaledhosny: While at it, can you make this a function and use it elsewhere in this file where we do:
```
if pos % 1 == 0:                                                    
    pos = int(pos)
```

python/psautohint/ufoFont.py
```diff
@@ -897,15 +900,22 @@ def getHintSet(self):
...
             hintset[STEMS_NAME].extend(
-                    makeHintSet(self.vList, self.vstem3List, isH=False))
+                makeHintSet(self.vList, self.vstem3List, isH=False))
 
         return hintset
 
 
+def norm_float(value):
+    """Converts a float (whose decimal part is zero) to integer"""
+    if isinstance(value, float) and value % 1 == 0:
```
anthrotype: `float.is_integer` is more idiomatic (not sure if faster)
https://docs.python.org/3/library/stdtypes.html#float.is_integer

miguelsousa:
```
~$ python3 -m timeit -s '(-2.0).is_integer()'
50000000 loops, best of 5: 7.93 nsec per loop
--------------------------------------------------------------------------------------------
~$ python3 -m timeit -s 'isinstance(-2.0, float) and -2.0 % 1 == 0'
50000000 loops, best of 5: 8.02 nsec per loop
```
`float.is_integer` is faster. I'll change it.

anthrotype: interesting. very small margin. I prefer is_interger mostly for legibility

miguelsousa: However, the incoming values are integers sometimes, and `(-2).is_integer()` yields an `AttributeError`, which means we need to have an extra check. Need to think about this some more.

anthrotype: yeah you still need to check isinstance(v, float) of course

miguelsousa: Right. So a better time comparison is this:
```
~$ python3 -m timeit -s 'isinstance(-2.0, float) and -2.0 % 1 == 0'
50000000 loops, best of 5: 8.03 nsec per loop
----------------------------------------------------------------------------------
~$ python3 -m timeit -s 'isinstance(-2.0, float) and (-2.0).is_integer()'
50000000 loops, best of 5: 8.02 nsec per loop
```

python/psautohint/ufoFont.py
```diff
@@ -911,7 +911,7 @@ def getHintSet(self):
 
 def norm_float(value):
     """Converts a float (whose decimal part is zero) to integer"""
-    if isinstance(value, float) and value % 1 == 0:
+    if isinstance(value, float) and value.is_integer:
```
khaledhosny: `is_integer` is a function.

anthrotype: a method yes. Ah.. the parentheses

miguelsousa: 🤦‍♂ moving too fast...

PR #181, opened 2019-08-09
Drop Python 2.7 support

### LINKED ISSUE #175: Drop support for Python 2.7
Synchronizing with recent updates to [AFDKO](https://github.com/adobe-type-tools/afdko/issues/741) and anticipating [Python 2.7's EOL as of January 1, 2020](https://pythonclock.org/), psautohint will only support Python 3.6+ moving forward. Code specific to supporting older Python will be removed.

- [x] update CI
- [x] remove `__future__` imports
- [x] remove use of `fontTools.misc.py23`
- [x] replace use of `.format()` with f-strings
- [x] update documentation
- [x] update `_psautohint.c` (remove `#if` conditional blocks for `PY_MAJOR_VERSION`)

josh-hadley:
Closes #175


tests/integration/differ.py
```diff
@@ -7,17 +7,13 @@
 Used as part of the integration tests.
```
khaledhosny: This script is copied from AFDKO, so I suggest updating it to the latest version on AFDKO repository.

tests/unittests/test_extension.py
```diff
@@ -57,8 +55,8 @@ def test_autohintmm_bad_args(args):
 
 
 @pytest.mark.parametrize("args", [
-    [(tounicode(GLYPH), GLYPH), (NAME, NAME)],  # 1st should be bytes
-    [(GLYPH, GLYPH), (tounicode(NAME), NAME)],  # 2nd should be bytes
+    [(GLYPH.decode('ascii'), GLYPH), (NAME, NAME)],  # 1st should be bytes
+    [(GLYPH, GLYPH), (NAME.decode('ascii'), NAME)],  # 2nd should be bytes
 ])
 @pytest.mark.skipif(sys.version_info < (3, 0),
```
khaledhosny: This `skipif` should go now.

tests/integration/differ.py
josh-hadley: Done

tests/unittests/test_extension.py
josh-hadley: Done

khaledhosny:
The C module should drop Python 2 support as well.

josh-hadley:
> The C module should drop Python 2 support as well.

@khaledhosny can you confirm that would only affect `_psautohint.c`?

khaledhosny:
> > The C module should drop Python 2 support as well.
> 
> @khaledhosny can you confirm that would only affect `_psautohint.c`?

Yes.

khaledhosny:
We might be also able to drop `libpsautohint/src/winstdint.h` now we don’t need to be compatible with the ancient MSVC required foe building with Python 2, but I’m OK if this was done later.

PR #192, opened 2019-09-25
[psstemhist] Adjust the formatting of the reports

### LINKED ISSUE #153: [psstemhist] beautify output files
current:
```
Vertical Stem List for /Users/fgriessh/Desktop/test/font.ufo on Mon Mar 25 18:29:45 2019
Count	Width	Glyph List
17	88	['a', 'd', 'four', 'g', 'h', 'i', 'j', 'l', 'm', 'n', 'one', 'p', 'q', 't', 'u']
11	93	['B', 'D', 'E', 'F', 'H', 'I', 'J', 'L', 'P', 'R', 'T']
11	87	['M', 'N', 'b', 'f', 'h', 'k', 'm', 'n', 'r']
3	94	['H', 'K', 'Y']
1	453	['seven']
1	441	['Z']
1	404	['Z']
1	388	['z']
1	360	['z']
1	339	['E']
1	338	['F']
1	236	['G']
1	190	['four']
1	92	['U']
1	91	['U']
1	90	['G']
```

suggested:
- get rid of Python-list structure
- right-align numbers
- use spaces instead of tabs for horizontal alignment
- sort glyph lists by glyph order
```
count    width    glyphs
   17       88    [a d g h i j l m n p q t u one four]
   11       93    [B D E F H I J L P R T]
   11       87    [M N b f h k m n r]
    3       94    [H K Y]
    1      453    [seven]
    1      441    [Z]
    1      404    [Z]
    1      388    [z]
    1      360    [z]
    1      339    [E]
    1      338    [F]
    1      236    [G]
    1      190    [four]
    1       92    [U]
    1       91    [U]
    1       90    [G]
```

even better:
- outliers removed 
```
count    width    glyphs
   17       88    [a d g h i j l m n p q t u one four]
   11       93    [B D E F H I J L P R T]
   11       87    [M N b f h k m n r]
    3       94    [H K Y]
    1       92    [U]
    1       91    [U]
    1       90    [G]
```

The same modifications would be desirable also for files written with option `-z`.

miguelsousa:
* get rid of Python-list structure
* right-align numbers
* use spaces instead of tabs for horizontal alignment
* sort glyph lists by glyph order

Fixes #153

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -154,14 +154,19 @@ class GlyphReports:
             bot_dict[bot] += 1
         return top_dict, bot_dict
 
-    @staticmethod
-    def assemble_rep_list(items_dict, count_dict):
+    def assemble_rep_list(self, items_dict, count_dict):
         # item 0: stem/zone count
         # item 1: stem width/zone height
         # item 2: list of glyph names
+        gorder = list(self.glyphs.keys())
         rep_list = []
         for item in items_dict:
-            rep_list.append((count_dict[item], item, sorted(items_dict[item])))
+            gnames = list(items_dict[item])
+            # sort the names by the font's glyph order
+            if len(gnames) > 1:
+                gindexes = [gorder.index(gname) for gname in gnames]
+                gnames = [x for _, x in sorted(zip(gindexes, gnames))]
+            rep_list.append((count_dict[item], item, gnames))
         return rep_list
 
     def _get_lists(self):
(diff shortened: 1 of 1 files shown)


miguelsousa:
I don't quite understand the failure. It seems that the top of the **t** (at 642) is not captured, and that gets replaced by a zone height at 486 which already existed. @khaledhosny does this make sense?

The new files were generated on my laptop (MacBook Pro, macOS 10.14.6, Python 3.7). I can tweak the results to make the CI pass, but it'd be better to understand the root of the problem.

```diff
--- /io/tests/integration/data/dummy/font.otf.top.txt
+++ /tmp/pytest-of-root/pytest-0/popen-gw0/test_stemhist_report_zones_0/font.otf.top.txt
@@ -14,3 +14,2 @@
    15      650    [zero two three six eight nine numbersign percent]
-    1      642    [t]
     8      638    [one four five seven]
@@ -21,3 +20,3 @@
     1      488    [W]
-   12      486    [f g i j t u v w x y z]
+   13      486    [f g i j t u v w x y z]
     2      468    [equal]
```

miguelsousa:
@cjchapman #193 seems to have fixed the inconsistencies. I'm now getting the same results locally.

python/psautohint/autohint.py
```diff
@@ -294,7 +297,8 @@ def save(self, path):
                 fp.write(header)
                 reps.sort(key=sortFunc)
                 for item in reps:
```
cjchapman: I realize this part of the code predates this change, but here at line 299 we have the start of a `for item in ...` loop that is *inside* another `for i, item in ...` loop starting at line 288. I suggest changing at least one of them so we're not using `item` in both places.

This is just a suggestion. Since it seems to work, and since it predates this PR, I'm going to go ahead and approve this PR.

miguelsousa: good catch. Changed.

PR #195, opened 2019-09-28
fix more memory issues

cjchapman:
this ~fixes~ is intended to fix the "mysterious crashes" mentioned in `test_stemhist.py`

### DIFF
--- a/libpsautohint/autohintexe.c
+++ b/libpsautohint/autohintexe.c
@@ -436,6 +436,7 @@ main(int argc, char* argv[])
             }
 
             ACBufferFree(output);
+            output=NULL;
             if (result != AC_Success)
                 exit(result);
         }
@@ -489,6 +490,7 @@ main(int argc, char* argv[])
             free(masters[i]);
             free(inGlyphs[i]);
             ACBufferFree(outGlyphs[i]);
+            outGlyphs[i] = NULL;
         }
         free(inGlyphs);
         free(outGlyphs);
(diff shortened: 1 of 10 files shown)


miguelsousa:
Even after these changes I still get `Abort trap: 6` when I run,

    pytest tests/integration/test_stemhist.py tests/unittests/test_extension.py

Interestingly, running each of those separately yields no problems.

    pytest tests/integration/test_stemhist.py
    pytest tests/unittests/test_extension.py

libpsautohint/autohintexe.c
```diff
@@ -436,6 +436,7 @@ main(int argc, char* argv[])
             }
 
             ACBufferFree(output);
+            output=NULL;
```
khaledhosny: It might be a good idea to make a, say, `AC_BUFFER_FREE` macro that calls `ACBufferFree()` and then do the `NULL` assignment, and use it everywhere `ACBufferFree()` is currently used.

cjchapman: I see can the advantages of that, but given the debugging pain I went through to get to this point, I'd really rather be able to search on pointers and see where they are set without having to remember what a macro does.

cjchapman: Also, just to be clear on what this point is: I think I've fixed the crashes, and now we just need to sort out the output differences. I've asked @readroberts to take a look at the output.

PR #197, opened 2019-10-01
CFF2 VF hinting

readroberts:
Refactored code to allow sharing paths between hinting source fonts for MM fonts, and hinting a CFF2 variable font. Hinting the master sources for a glyph for the two cases is now identical. The latter differs from the former in that the VF charstring must first be interpolated to a regular T2 charstring for each master source font. After the hinting is done, the T2 charstrings must be merged back together into a single VF charstring.

I also renamed some variables that had different meanings for the same name. For example, 'width' in some contexts contains the advance width for a glyph. In other contexts, it contains the T2 charstring width arg, which is (advance width - CFF nominal width). I renamed all occurrences of 'width' in the latter context to 't2_width_arg'.

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -29,15 +29,13 @@
 #  Save font plist file.
 
 import ast
-import copy
 import logging
 import os
 import re
 import time
 from collections import defaultdict, namedtuple
 
-from .otfFont import (CFFFontData, convertT2GlyphToBez, convertBezToT2,
-                      interpolate_cff2_charstring)
+from .otfFont import (CFFFontData)
 
 from .ufoFont import UFOFontData
 from ._psautohint import error as PsAutoHintCError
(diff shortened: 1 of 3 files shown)


miguelsousa:
Coverage went down quite a bit. See https://codecov.io/gh/adobe-type-tools/psautohint/pull/197/changes

Major changes like these must be accompanied by thorough test cases.

readroberts:
I already have test cases, and will be adding them soon. I am waiting until I have finished more extensive testing to see if I would need to add more test cases.

khaledhosny:
> I also renamed some variables that had different meanings for the same name. For example, 'width' in some contexts contains the advance width for a glyph. In other contexts, it contains the T2 charstring width arg, which is (advance width - CFF nominal width). I renamed all occurrences of 'width' in the latter context to 't2_width_arg'.

The T2 width is not really used modified, it is passed around to be used by `updateFromBez()`. `CFFFontData` could very well just cache when extracting the CharString to be used later and stop passing it around.

python/psautohint/autohint.py
```diff
@@ -29,23 +29,49 @@
...
 import logging
 import os
 import re
 import time
 from collections import defaultdict, namedtuple
 
-from .otfFont import CFFFontData
+from .otfFont import (CFFFontData, convertT2GlyphToBez, convertBezToT2,
+                      interpolate_cff2_charstring)
+
```
khaledhosny: The current direction of this file is to keep it neutral to the font format, and encapsulate the glyph parsing and updating logic in the respective font classes. It would be really nice to keep it that way if it is not too much effort.

readroberts: Sure.

python/psautohint/ufoFont.py
```diff
@@ -382,17 +382,21 @@ def getPSName(self):
...
 
         bezString = "\n".join(bez)
         bezString = "\n".join(["% " + name, "sc", bezString, "ed", ""])
-        return bezString, width
+        return bezString
 
-    def updateFromBez(self, bezData, name, width, mm_hint_info=None):
+    def updateFromBez(self, bezData, name, mm_hint_info=None):
         # For UFO font, we don't use the width parameter:
         # it is carried over from the input glif file.
```
khaledhosny: This comment is outdated now.

PR #200, opened 2019-10-04
Unhack hint_compatible_glyphs

miguelsousa:


python/psautohint/autohint.py
```diff
@@ -587,7 +587,19 @@ def hint_glyph(options, name, bez_glyph, fontinfo):
 
 def hint_compatible_glyphs(options, name, bez_glyphs, masters, fontinfo):
     try:
-        hinted = hint_compatible_bez_glyphs(fontinfo, bez_glyphs, masters)
```
khaledhosny: I think what is needed here is to filter out `None` glyphs and replace them with empty ones (`# name\ncp`, IIRC) or something like that.

The idea of calling `hint_compatible_bez_glyphs()` with all masters at once is to avoid hinting the first master over and over again for every other master.

josh-hadley: I tried both of those approaches while debugging this with no luck (it either breaks the C side of the code, which is apparently not set up to deal with sparse masters, or defers related problems in the Python code until later in the process).

I think maybe we should just close this PR and come back to the larger issue of sparse master support some other time. @miguelsousa thoughts?

miguelsousa: I defer to Khaled. He knows better how things were meant to work. On the surface it does seem like something got overlooked when support for sparse masters was implemented. The comment may be stale as well.
I'm fine with closing the PR and postpone the fix, whatever that might be.

josh-hadley:
Decided to close for now, and revisit when we have time to make thorough and complete changes to address the root issue.

PR #203, opened 2019-10-17
Refactored code to allow sharing paths between hinting source fonts f…

readroberts:
…or MM fonts, and hinting a CFF2 variable font. Hinting the master sources for a glyph for the two cases is now identical. The latter differs from the former in that the VF charstring must first be interpolated to a regular T2 charstring for each master source font. After the hinting is done, the T2 charstrings must be merged back together into a single VF charstring.

I also renamed some variables that had different meanings for the same name. For example, 'width' in some contexts contains the advance width for a glyph. In other contexts, it contains the T2 charstring width arg, which is (advance width - CFF nominal width). I renamed all occurrences of 'width' in the latter context to 't2_width_arg'.

Refactored code in autohint.py and otffont.py to move font-format specific calls and properties into otffont.py, per Khaled's review.

Fixed new LGTM issues.

### DIFF
--- a/python/psautohint/ufoFont.py
+++ b/python/psautohint/ufoFont.py
@@ -382,19 +382,21 @@ class UFOFontData:
     def isCID():
         return False
 
+    @staticmethod
+    def hasFDArray():
+        return False
+
     def convertToBez(self, name, read_hints, round_coords, doAll=False):
         # We do not yet support reading hints, so read_hints is ignored.
         width, bez, skip = self._get_or_skip_glyph(name, round_coords, doAll)
         if skip:
-            return None, width
+            return None
 
         bezString = "\n".join(bez)
         bezString = "\n".join(["% " + name, "sc", bezString, "ed", ""])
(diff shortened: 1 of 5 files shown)


miguelsousa:
Coverage went down https://codecov.io/gh/adobe-type-tools/psautohint/pull/203/diff

python/psautohint/ufoFont.py
```diff
@@ -382,17 +382,21 @@ def getPSName(self):
...
 
         bezString = "\n".join(bez)
         bezString = "\n".join(["% " + name, "sc", bezString, "ed", ""])
-        return bezString, width
+        return bezString
 
-    def updateFromBez(self, bezData, name, width, mm_hint_info=None):
+    def updateFromBez(self, bezData, name, mm_hint_info=None):
         # For UFO font, we don't use the width parameter:
         # it is carried over from the input glif file.
```
khaledhosny: This comment should be dropped now.

PR #211, opened 2019-11-20
added "g_" prefix to global names to resolve conflicts with local names

cjchapman:


khaledhosny:
That is all fine, but IMO it is still good to be able to tell the scope of the variable from its name (that why I renamed all these global variables and gave them the `g` prefix in the first place). But I’m not actively touching this code so I’ll leave it to you to decide.

cjchapman:
OK.

PR #214, opened 2019-11-21
remove the global variable `g_links`

cjchapman:


libpsautohint/src/shuffle.c
```diff
@@ -30,12 +30,13 @@ InitShuffleSubpaths(void)
     }
     cnt++;
     rowcnt = cnt;
-    g_links =
+    links =
       (cnt < 4 || cnt >= MAXCNT) ? NULL : (unsigned char*)Alloc(cnt * cnt);
+    return links;
```
khaledhosny: You can do away with the variable and return directly. Even more readable:
```C
if (cnt < 4 || cnt >= MAXCNT)
  return NULL;
return Alloc(cnt * cnt);
```

cjchapman: Done. Thanks.

PR #216, opened 2019-11-21
Address remaining LGTM.com alerts (Python)

josh-hadley:
4 warnings
4 recommendations

This _should_ be everything, for the currently-identified alerts by LGTM.com

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -620,28 +620,31 @@ def hint_compatible_glyphs(options, name, bez_glyphs, masters, fontinfo):
     # and hint_vf_font.
     try:
         ref_master = masters[0]
-        if False:
-            # This is disabled because it causes crashes on the CI servers
-            # which are not reproducible locally. The branch below is a hack to
-            # avoid the crash and should be dropped once the crash is fixed,
-            # https://github.com/adobe-type-tools/psautohint/pull/131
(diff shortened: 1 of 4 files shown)


python/psautohint/autohint.py
```diff
@@ -620,28 +620,31 @@ def hint_compatible_glyphs(options, name, bez_glyphs, masters, fontinfo):
...
-                else:
-                    hinted.append(out[1])
+        # if False:
+        #     # This is disabled because it causes crashes on the CI servers
+        #     # which are not reproducible locally. The branch below is a hack
+        #     # to avoid the crash and should be dropped once the crash is
+        #     # fixed, https://github.com/adobe-type-tools/psautohint/pull/131
+        #     hinted = hint_compatible_bez_glyphs(
+        #         fontinfo, bez_glyphs, masters)
+        # *** see https://github.com/adobe-type-tools/psautohint/issues/202 ***
```
cjchapman: Please move the `see http:...` comment to the top of this block along with something like "DO NOT DELETE THIS COMMENTED OUT CODE" and generally make it more visible so some well-meaning person doesn't remove this during a code cleanup. Thanks.

josh-hadley: fixed in https://github.com/adobe-type-tools/psautohint/pull/216/commits/fdf96d64fe5b74516de2763e313b62f8ab69128e

cjchapman: 👍

Issue #223, opened 2020-01-14
-x fails with maximum recursion error


kontur:
Hey,

I'm trying to autohint a font, but hit a hinting error on a particular glyph. I am trying to exclude that particular glyph with `-x`, but always hit another error relating to the`-x` param. Using `-g` on the same glyph (to verify syntax / CID accessing) as expected I get the same initial hinting error.

```
$ psautohint -vv Font.otf
...hinting
DEBUG: u043B.smcp: 54.5  u043B.smcp                    439.5  494
ERROR: u043B.smcp:   Conflicts with current hints: 0 -6 0 76 b.
ERROR: u043B.smcp: Uninitialized extra hints list.
ERROR: u043B.smcp: Failure in processing outline data.

$ psautohint -vv -x u043B.smcp Font.otf
DEBUG: Reading 'CFF ' table from disk
DEBUG: Decompiling 'CFF ' table
ERROR: maximum recursion depth exceeded

$ psautohint -vv -g u043B.smcp Font.otf
...hinting
DEBUG: u043B.smcp: 54.5  u043B.smcp                    439.5  494
ERROR: u043B.smcp:   Conflicts with current hints: 0 -6 0 76 b.
ERROR: u043B.smcp: Uninitialized extra hints list.
ERROR: u043B.smcp: Failure in processing outline data.

What's causing the -x to fail?

**Comment by kontur** [comment]
What's more, `--exclude-glyphs-file` seems to first interpret the file contents as string arrays instead of glyph names, but then fails with the same error:

```
$ cat excl.txt
u043B.smcp
$ psautohint -vv --exclude-glyphs-file excl.txt Font.otf
DEBUG: Reading 'CFF ' table from disk
DEBUG: Decompiling 'CFF ' table
WARNING: glyph ID <u> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <0> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <4> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <3> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <B> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <.> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <s> from glyph selection list option is not in font. <Font.otf>.
WARNING: glyph ID <m> from glyph selection list option is not in font. <Font.otf>.
[...]

PR #224, opened 2020-01-14
Fixed for recursion error on -x #223

kontur:
Tried my hand at a fix for the recursion error by rewriting the lambda to a list comprehension.

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -362,8 +362,7 @@ def filterGlyphList(options, fontGlyphList, fontFileName):
             if glyphNames is not None:
                 glyphList.extend(glyphNames)
         if options.excludeGlyphList:
-            newList = filter(lambda name: name not in glyphList, fontGlyphList)
-            glyphList = newList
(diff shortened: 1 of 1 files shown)


josh-hadley:
Looking at the failures in CI, it appears to be something with `coverage` ... tests are all passing in the `tox` run; the failures happen when `coverage` attempts to read the files. There were some recent updates to `coverage`; I'll poke around on this a bit and see if I can get things sorted out and get these tests passing and make some progress on this PR.

josh-hadley:
Bad day for Python/pip-based CI, apparently. Travis is failing, looks like something in the multibuild dependency (same error as mentioned in https://github.com/matthew-brett/multibuild/issues/300) which I am guessing is likely related to a recent spate of problems with pip & packaging today.

But, in good news: the update I made to `tox.ini` gets AppVeyor CI tests to pass. I'm going to just hold for a while and see if there are updates to any of the dependencies and retry when things get resolved. I'm reasonably confident that your code is OK.

josh-hadley:
TravisCI issues *finally* sorted out; approved and will go ahead and merge! Thanks for your contribution @kontur!

Issue #222, opened 2020-01-14
--print-dflt-fddict results in error


kontur:
It seems any font I try the `--print-dftl-fddict` param on I get an error of `ERROR: 'NoneType' object is not subscriptable`.

I haven't debugged this any further, but since three different font files I tried all resulted in this categorical error there seems to be something amiss with the values to report not getting passed.

kontur:
Firstly, this [return](https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/autohint.py#L532) seems to prevent returning a `fontinfo_list`, which then bubbles up and ultimately fails the call.

Is calling with `--print-dftl-fddict` intended to end execution after the print of that information? If so, then it needs to exit more gracefully.

Secondly, this [try with silent pass](https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/autohint.py#L423) is the reason there is no actual output, despite the passed in  `fontInfoString` actually having content. The whole `printFontInfo` seems pretty funky... a string representation of FDDict is passed to it, then some keys are matched with those regex above.

Not knowing what the formatting of this output here used to be, I wonder if those calls to `printFontInfo(str(fontDict))` shouldn't simply be replaced with something like `print(fontDict.getFontInfo())` (which [FDDict already has](https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/fdTools.py#L114), or better yet give `FDDict` a proper `__repr__` method that does what it should here and simply `print(fontDict)`?

Happy to help with this, but need a bit insight as to what this bit of code is expected to return and how.

kontur:
Gave it a fix attempt, anyway.

josh-hadley:
Thanks for your attempt on this. I agree with your assessment that the empty `return` interferes with the subsequent operations that expect a `fontinfo_list`, and the try-with-silent-pass hides this (and other problems).
I'll have to check in with other folks as to the intended behavior of `--print-dflt-fddict`. It does seem to me like the intent was to simply print and exit. Might be some clues in the now-defunct AFDKO autohint.py program that psautohint was based on. I'll do some research.

josh-hadley:
Looks like this option has been broken for a while :-(
At any rate: I went back to an older AFDKO, prior to deprecating the old autohint.py program. From the behavior of that, it definitely looks like the intent with `--print-dflt-fddict` (which was `-pfd` option in autohint.py) was to print then exit.

kontur:
[...]

PR #225, opened 2020-01-14
Simplifying code to print the FDDict with --print-dflt-fddict and exiting

kontur:
Potential fix for #222

I'm not sure should or should--print-dflt-fddict exit or not, but with this fix it does. When omitting [this return statement](https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/autohint.py#L532) the code continues to execute, but then a loooot of CLI spam follows, so presumably this used to exit after the print of the dflt values.

The printing of the FDDict seems to have been formatted with tabs, whereas the existing `getFontInfo` [method](https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/fdTools.py#L114) seems to not - but without further knowledge it is hard to say if this implementation is usable or if some other kind of formatting is required.

### DIFF
--- a/python/psautohint/autohint.py
+++ b/python/psautohint/autohint.py
@@ -31,7 +31,7 @@
 import ast
 import logging
 import os
-import re
+import sys
 import time
 from collections import defaultdict, namedtuple
 
@@ -402,28 +402,6 @@ fontInfoKeywordList = [
     'Baseline6',
 ]
 
-integerPattern = r""" -?\d+"""
-arrayPattern = r""" \[[ ,0-9]+\]"""
-stringPattern = r""" \S+"""
-counterPattern = r""" \([\S ]+\)"""
-
-
-def printFontInfo(fontInfoString):
-    for item in fontInfoKeywordList:
-        if item in ['FontName', 'FlexOK']:
-            matchingExp = item + stringPattern
-        elif item in ['VCounterChars', 'HCounterChars']:
-            matchingExp = item + counterPattern
-        elif item in ['DominantV', 'DominantH']:
-            matchingExp = item + arrayPattern
-        else:
-            matchingExp = item + integerPattern
-
-        try:
-            print('\t%s' % re.search(matchingExp, fontInfoString).group())
-        except Exception:
-            pass
-
 
 def openFile(path, options):
     font_format = get_font_format(path)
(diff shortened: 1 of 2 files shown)


kontur:
How should this CI fails be interpreted? I am not familiar with that.

josh-hadley:
> How should this CI fails be interpreted? I am not familiar with that.

There are a couple of issues: one is related to our code coverage testing; basically there's a newer version of the coverage tool and a mismatch in our CI configuration. I've addressed that with https://github.com/adobe-type-tools/psautohint/commit/35b7cf49944e9185e0915caf6ce545cddc3614ee . The other issue is related to a host of problems with `pip` and other packaging components that came up last week and threw a spanner into the works of a _lot_ of people using Python in CI. I _think_ that is resolved now, but I'm not sure, and will test further this week.

josh-hadley:
I did some research and found that an old version of `autohint` (from AFDKO) yields the following output for the command `autohint -pfd font.otf` (equivalent of `psautohint font.otf --print-dflt-fddict`):
```
Showing default FDDict Values:
	FontName SourceCodePro-Regular
	OrigEmSqUnits 1000
	LanguageGroup 0
	DominantV [86]
	DominantH [68]
	FlexOK true
	BlueFuzz 0
	BaselineYCoord 0
	BaselineOvershoot -12
	CapHeight 486
	CapOvershoot 12
	LcHeight 574
	LcOvershoot 12
	AscenderHeight 638
	AscenderOvershoot 12
	FigHeight 656
	FigOvershoot 12
	Height5 712
	Height5Overshoot 12
	Baseline5Overshoot -12
	Baseline5 -205
```
The output from this PR is close, but lacks the tabbed indentation (which, I gathered from further research is probably there to facilitate adding the output of this command into a `fontinfo` file without additional text manipulation).

Additionally: now that we have the CI failures sorted out 🤞, I was able to get to the bottom of the test failure: it seems the test for this command lumps it together with several other tests, and I think the inputs/options for those are not appropriate for this command, and generate unrelated errors. So I think the test needs to change for this case (and we can use the output from the old `autohint` as the expected data).

@kontur : can you make the code changes needed to make this PR do tab-indented formatting? If you can do that, I will update the test code and data and we can get this PR closed out and merged.

kontur:
Output indented now, but the CI still throws some meta errors:

> An object at the specified path .tox\\log\\*.log does not exist, or has been filtered by the -Include or -Exclude parameter.

josh-hadley:
> Output indented now

Thanks!

> but the CI still throws some meta errors

Might be related to the test failure. I'll get the test code & data updated on this PR and see if we can get it working.

josh-hadley:
Yay! With updated test data & test code, everything's passing, so I'll merge now. @kontur thanks again for your patience getting this one through!

kontur:
Cool, glad to be contributing a small fix as thanks of many hours saved not having to hint ;)

Issue #231, opened 2020-02-17
PFA: tx reports unhinted glyphs


frankrolf:
Minor issue, but possibly confusing:

When hinting a pfa file via `psautohint font.pfa`, `tx` reports the unhinted status of the source file. I understand `tx` is used to convert from the pfa format to cff (#81) – perhaps the `tx` “unhinted” output messages could be suppressed here?

```
psautohint font.pfa
tx: --- /Users/fgriessh/Desktop/font.pfa
tx: (cfw) unhinted <.notdef>
tx: (cfw) unhinted <exclam>
tx: (cfw) unhinted <quotedbl>
tx: (cfw) unhinted <numbersign>
tx: (cfw) unhinted <dollar>
tx: (cfw) There are 269 additional reports of 'unhinted'.
```

josh-hadley:
Are you thinking that `psautohint` should do the filtering? I will need to look into it a bit more but I think that should be possible (scan for that pattern in the captured output from `tx` when it is called) and would probably be easier than modifying `tx` to suppress those messages.

frankrolf:
I have no preference how (and if) the filtering should be done. All I’d like to say is that the output here might be slightly confusing.

PR #261, opened 2020-06-25
Mute 'unhinted <glyphname>' messages from tx

miguelsousa:
Relates to #231

Adding a test to confirm the messages first, then follow with the actual fix.


miguelsousa:
The added test ends up being skipped by the CI because it requires `tx`. Not great but better than nothing, I guess.

PR #272, opened 2020-11-06
flake8 fixes

### LINKED ISSUE #207: Add flake8 as gating check
Add a flake8 check to be used in CI such that if the flake8 test fails, other tests will be skipped.

~Note that the behavior of `tox` is to run everything in `envlist`, whether there are failures or not. So, the flake8 test should be added as follows:~

~- [ ] define a `[testenv:flake8]` with the flake8 `commands` and `deps`, but do _not_ add `flake8` to the `[tox] envlist`~
~- [ ] invoke in CI as something like `tox -e flake8 && tox` such a flake8 failure short-circuits and the other tests won't waste resources running the other environments/tests~

**Update Nov 6, 2020:**
Changes in TravisCI have accelerated a planned switch to GitHub Actions for testing & deployment. That switch is now complete. As a result of changes to the testing scheme, the task list is slightly different:
- [x] define a `[flake8]` section in `setup.cfg` using a similar scheme as is used in [AFDKO's `.flake8` config](https://github.com/adobe-type-tools/afdko/blob/develop/.flake8)
- [x] update the [psautohint testing workflow](https://github.com/adobe-type-tools/psautohint/blob/master/.github/workflows/pythonpackage.yml) to simply run `flake8` using that configuration
- [ ] update any psautohint code to address any ignored errors or warnings, and remove those from the config's ignore list.

josh-hadley:
- remove temporarily ignored `flake8` checks
- clean up Python code to address those checks

closes #207

### DIFF
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -871,7 +871,7 @@ class CFFFontData:
         if font_format == "OTF":
             # It is an OTF font, we can process it directly.
             font = TTFont(path)
-            if "CFF "in font:
+            if "CFF " in font:
                 cff_format = "CFF "
             elif "CFF2" in font:
                 cff_format = "CFF2"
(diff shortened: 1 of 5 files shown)


cjchapman (approved):
👍

PR #187, opened 2019-08-15
Update psautohint to use autohint CLI options

josh-hadley:
Use legacy `autohint` CLI flags & fix "intuitive" use of `-o` option (output preceding input).

closes #176 
closes #129

PR #281, opened 2020-12-02
Remove support for legacy-named options

### LINKED ISSUE #260: Deprecation/compatibility option not shown for some options
In [an AFDKO review relating to psautohint](https://github.com/adobe-type-tools/afdko/pull/1175#discussion_r445779690) it was noted that a deprecation message does not appear when the legacy `-hfd` flag is used. The deprecation warning does appear for other legacy options, e.g. `-pfdl`.

Cause appears to be in https://github.com/adobe-type-tools/psautohint/blob/d1ca70affebe4f070037365d44780ab8a78778b0/python/psautohint/__main__.py#L829-L835
The `action` does not use `_DeprecatedAction` like other deprecated/legacy compatibility options do.

miguelsousa:
Support for legacy-named options was added in #187. It has been more than a year, so it's time to move on.
Incidentally it also fixes #260.


josh-hadley (approved):
👍  Thank you!

Issue #221, opened 2020-01-13
Performance on variable fonts?


madig:
Hinting Cantarell-VF takes a long time on my Ryzen 1700. I used https://github.com/benfred/py-spy to profile the run and had it generate a speedscope file: [profile.zip](https://github.com/adobe-type-tools/psautohint/files/4055894/profile.zip). The file can be loaded on https://www.speedscope.app/.

Something like 90% of the time is spent in `SimpleT2Decompiler.execute`, which itself spends most of that time in `T2CharString.getToken`. If performance gets on the todo list, this is probably a good place to start looking.

josh-hadley:
Just noting that these routines are actually in the fonttools codebase: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/misc/psCharStrings.py

madig:
Wherein I look at a debugger and scratch my head: so I put a `logging.warning` before and after https://github.com/fonttools/fonttools/blob/4.18.2/Lib/fontTools/subset/cff.py#L492 because the profiler says 90% of time is spent in there and it dutifully prints me a list of processed glyph names. And then it keeps going for a long time after the end, without triggering any more logging messages. Ctrl+C-ing at some point does show exactly that line in the backtrace though and going up the stack frames reveals that the same glyph names that were processed the first time around show up again. Huh. Almost as if the same glyphs are desubroutinized over and over again, but somehow without triggering my loggers.

PR #289, opened 2021-01-01
CFF2: Desubroutinize once, not again for each glyph

madig:
Closes https://github.com/adobe-type-tools/psautohint/issues/221.

Performance on VFs was dog-slow because the entire font was desubroutinized anew for each glyph, so psautohint was [accidentally quadratic](https://accidentallyquadratic.tumblr.com/) I guess?

### DIFF
--- a/python/psautohint/otfFont.py
+++ b/python/psautohint/otfFont.py
@@ -906,11 +906,16 @@ class CFFFontData:
         self.topDict = self.cffTable.cff.topDictIndex[0]
         self.charStrings = self.topDict.CharStrings
         if 'fvar' in self.ttFont:
+            # have not yet collected VF global data.
             self.is_vf = True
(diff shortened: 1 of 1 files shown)


punchcutter:
Thanks for finding this. I ran into the same problem a few weeks ago and we didn't have time to check it before the holidays.

punchcutter:
Nice, I ran it on this 65k glyph CJK font and it finished in about 8 minutes instead of before where it ran overnight for 12 hours and only did 1000 glyphs and then crashed.

madig:
There's further scope for optimization. The profiler traces point out a lot of (useless?) action going on in logging functions, also in the C extension if I'm not mistaken.

josh-hadley (approved):
🎆  LGTM 🎉 
@madig thanks so much for delving into this and fixing it!

PR #301, opened 2021-04-20
Keep up with fontTools getDeltas API

madig:
Using the current fontTools will produces errors on VFs about `getDeltas` getting too many arguments. See https://github.com/fonttools/fonttools/pull/2214.


madig:
Oh, right. 4.22.0 also changes TTX dumps so everything has to be updated now.

madig:
Needs https://github.com/adobe-type-tools/psautohint-testdata/pull/2.

PR #317, opened 2021-07-13
Corrected/better flex heuristics

iterumllc:
TryXFlex was slightly mangled by https://github.com/adobe-type-tools/psautohint/commit/c951c5df66ff42c96ce0e7d89656e3aae347ab3a and this PR fixes that.

It also changes the x/y ratio heuristic. The flexdiff is between p0 or p2 and p1, not between p0 and p2, as should be evident by the "near miss" heuristic at the end. This uses the average of p0 and p2 (in the relevant direction) compared with p1.


iterumllc:
I didn't change the Libertinus integration test as it is a good indication of what this (now draft) PR will do. I can do that if and when the code is approved of.

iterumllc:
This is no longer needed.

PR #338, opened 2021-12-13
Drop Python 3.6 support

miguelsousa:
These changes,
- Make Python 3.7+ the minimum supported version
- Drop Python 3.6 wheels
- Add `arm64` wheels for macOS
- Make the `build_wheels` job into a separate GH workflow
- Make various updates and enhancements to all GH workflow files
- Update `lxml` and `fonttools` to the latest versions

I also tried to add testing and wheel-building for Python 3.10, but ran into `tox` shenanigans, so decided to bail on that. The wheels probably build just fine (didn't confirm though), but didn't think it was a good idea to add them to the release without having a guarantee that the tests pass on Python 3.10.


miguelsousa:
The **build_wheels** job was not supposed to run, so I canceled it.

josh-hadley (approved):
LGTM, thanks for the updates. We'll investigate the `tox` setup more and try to figure out what's going on there with 3.10.

PR #344, opened 2021-12-26
make otfFont.py not invoke tx with `-std`

tkacvins:
Simple change to make psautohint not assume `StandardEncoding`


tkacvins (commented):
This won't work if the input `PFA` `StandardEncoding`.  The resulting `PFA` from running `psautohint` has an `/Encoding` of pure `/.notdef`s

tkacvins:
Fixes an edge case, but does not take into account the general cases

PR #348, opened 2022-01-17
Remove tox dependency. Add Python 3.10 support

miguelsousa:
By removing **tox** it makes it more onerous for a developer to run the tests on a multitude of Python environments locally. On the other hand, removing tox streamlines our test-config setup, and makes it easier (and faster!) to run the tests on different Python+OS configurations through the GitHub Actions CI.

The CI workflow now runs faster ([9min](https://github.com/adobe-type-tools/psautohint/actions/runs/1706614735) vs. [14min](https://github.com/adobe-type-tools/psautohint/actions/runs/1698661828)) because code coverage was always on.

The Python 3.10 wheels are built in a separate step from the Python 3.7—3.9 wheels because [the `manylinux1` image does not support CPython 3.10](https://github.com/pypa/manylinux/issues/994).


alerque:
Ditching tox is a great thing for distro packagers. Upstream suites that cater to tox and/or other isolation techniques make it really hard to test code in the context of a distribution package — which for us is what matters.

alerque (approved):

josh-hadley (approved):
Thank you @miguelsousa !

Issue #302, opened 2021-04-26
flattenBlends should test on empty incoming list


typemytype:
blue values or V/H stem list could be empty

https://github.com/adobe-type-tools/psautohint/blob/master/python/psautohint/otfFont.py#L1015

```traceback
...
python3.7/psautohint/otfFont.py", line 1015, in flattenBlends
IndexError: list index out of range
```

typemytype:
this happens when fe blues are set but no otherBlues

iterumllc:
This should be fixed in the 'pyport' branch.

frankrolf:
It’s hard to find that branch without a link @skef ;-)

skef:
Refactored in otfautohint in afkdo

PR #395, opened 2022-11-25
simple test if incoming list is not empty

typemytype:
fixing #302


frankrolf:
Frederik and I had a discussion about this – while we understand that many bugs are fixed in https://github.com/iterumllc/psautohint/tree/pyport, some small bugfixes could still make it into the main branch here (which is embedded in RoboFont)

skef:
Seems fine to me.

Any chance of moving some of the booleanOperations stuff forward while we're looking at things?

skef (approved):
lgtm

skef:
This was never merged and has been superseded by otfautohint in the main afdko repository.

